PR #11, opened 2022-02-25
Load tester to use simulcast layers

davidzhao:


pkg/loadtester/loadtester.go
```diff
@@ -140,29 +141,47 @@ func (t *LoadTester) PublishTrack(name string, kind lksdk.TrackKind, bitrate uin
...
+		fmt.Println("using simulcast quality ", i, "bitrate", layer.Bitrate)
+
+		sampleProvider, err := NewLoadTestProvider(layer.Bitrate)
+		if err != nil {
+			return "", err
+		}
+
+		track, err := lksdk.NewLocalSampleTrack(webrtc.RTPCodecCapability{
+			MimeType:    webrtc.MimeTypeVP8,
+			ClockRate:   33,
```
boks1971, line 166: Why is the clock rate 33 here and not 90K?

boks1971: Looks like we have to use frame interval in ms. Guess that is how that thing works. I see audio also set to 20.

boks1971 (approved):

boks1971:
@davidzhao  Maybe worth adding a non-simulcast video track too (would be even better if we have a screen sharing kind of stream with bitrates spiking up periodically). Maybe in the code that sets up a video, you can just draw a random number and if that is less than 0.1, set up a non-simulcast track to give 10% of the tracks be non-simulcast in the test.

matkam (approved):

Issue #27, opened 2022-05-23
Error when running livekit-load-tester on v0.8.0


stevenhansel:
I was trying to run this command:
```
livekit-load-tester \
       --url wss://wss.example.com \
       --api-key my_api_key \
       --api-secret my_api_secret \
       --duration 1m \
       --publishers 1 \
       --subscribers 5
```

Then, I was getting this error:

```
Starting load test with 1 publishers, 5 subscribers, room:
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x6 pc=0xa8172e]

goroutine 572 [running]:
github.com/livekit/livekit-cli/pkg/provider.(*H264VideoLooper).nextSample(0xc000335b00, 0x1)
        /root/go/pkg/mod/github.com/livekit/livekit-cli@v0.8.0/pkg/provider/h264looper.go:77 +0x38e
github.com/livekit/livekit-cli/pkg/provider.(*H264VideoLooper).NextSample(0xc000335b00?)
        /root/go/pkg/mod/github.com/livekit/livekit-cli@v0.8.0/pkg/provider/h264looper.go:54 +0x48
github.com/livekit/server-sdk-go.(*LocalSampleTrack).writeWorker(0xc0003cfad0, {0x7fcfd9be30c0, 0xc000335b00}, 0x0)
        /root/go/pkg/mod/github.com/livekit/server-sdk-go@v0.10.0/localsampletrack.go:395 +0x1c7
created by github.com/livekit/server-sdk-go.(*LocalSampleTrack).Bind
        /root/go/pkg/mod/github.com/livekit/server-sdk-go@v0.10.0/localsampletrack.go:185 +0x57b
```

My current solution is to just downgrade to v0.7.2
`go install github.com/livekit/livekit-cli/cmd/livekit-load-tester@v0.7.2`

davidzhao:
@stevenhansel

I believe this is due to interaction between `go install` and Git LFS. For now, can you try the following:

1. ensure git lfs is installed
2. clone the repo
3. run `make install`

stevenhansel:
I just tried installing git lfs, and running make install but it seems that the error still persists. I actually tried debugging the CLI and here's what I found:

I was curious to see what the error contains in `h264looper.go`

```
func (l *H264VideoLooper) nextSample(rewindEOF bool) (media.Sample, error) {
	sample := media.Sample{}
	if l.reader == nil {
		var err error
[...]

PR #28, opened 2022-05-23
Update release.yml to dereference files from Git LFS

stevenhansel:
This PR intends to fix issue #27 where an error occured in v0.8.0 because the build doesn't pull the LFS files in `pkg/provider/resources`.

In order to solve this problem, we just have to tell the workflow in `.github/workflows/release.yml` to download the LFS files so it's included in the build. But, there's a currently ongoing issue with the LFS that you can see [here](https://github.com/actions/checkout/issues/270).

You can see the sample release files from the updated release.yml [in my fork](https://github.com/stevenhansel/livekit-cli/releases/tag/v0.8.1)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,8 +12,15 @@ jobs:
     name: Generate cross-platform builds
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout the repository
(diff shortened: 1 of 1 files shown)


davidzhao:
This is a great find Steven! Merging!

PR #36, opened 2022-07-01
fix version ref

chenrui333:


davidzhao (approved):
thanks for fixing the oversight!

chenrui333:
NP

chenrui333:
just signed the CLA.

davidzhao:
I'll re-tag and re-release 1.0.1. will update the brew PR after that.

PR #37, opened 2022-07-01
deps: update to use go1.18

chenrui333:
also update to use go1.18 (match with [the formula setup](https://github.com/Homebrew/homebrew-core/pull/104900))


chenrui333:
need [this commit](https://github.com/dominikh/go-tools/commit/9a360a181e0c0b16c3ca971e99df6d37562f3b01) to be released on the checkstatic side

PR #62, opened 2022-07-14
publishFile: exit(0) after finished writing file

codenoid:
currently `publishFile` just stuck after finish writing a file

CMIIW


davidzhao (changes requested):
Hi there, thanks for the contribution.

I think this would be a good mode to add to the CLI, but it should be behind a flag (maybe `--exit-after-publish`)

It would also be ideal not to run os.Exit(0), but to run its clean up sequences so the participant could cleanly disconnect from the room.

PR #67, opened 2022-07-22
Support TCP sockets in addition to Unix sockets

bsirang:
Adds support for TCP sockets.

Using a new format for the "name" string passed in via `--publish`

The format is `mime_type://address:port` in the case of TCP or `mime_type:://unix_socket/path` in the case of unix sockets

The existence of the second `:` in the string is used to assume a TCP endpoint in the address:port format.

### DIFF
--- a/cmd/livekit-cli/join.go
+++ b/cmd/livekit-cli/join.go
@@ -110,11 +110,13 @@ func joinRoom(c *cli.Context) error {
 
 func handlePublish(room *lksdk.Room, name string, fps float64) error {
 	// See if we're dealing with a socket
-	mime_type, socket_type, address, err := parseSocketFromName(name)
-	if (err == nil) {
+	if isSocketFormat(name) {
+		mime_type, socket_type, address, err := parseSocketFromName(name)
+		if (err != nil) {
+			return err
+		}
 		return publishSocket(room, mime_type, socket_type, address, fps)
 	}
-
 	// Else, handle file
 	return publishFile(room, name, fps)
 }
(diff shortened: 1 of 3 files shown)


dwirya:
Hi @bsirang, nice work! If I may suggest, the URL scheme `tcp://{host}:{port}/{mime_type}` looks cleaner than using "::" to differentiate between TCP and Unix sockets. We can also drop support for Unix sockets in favour of TCP so that it's not confusing to users. I'll let @davidzhao decide on the best path forward.

bsirang:
Got it. I'll wait to hear back from @davidzhao before making any more changes here

cmd/livekit-cli/join.go
```diff
@@ -175,26 +178,57 @@ func publishFile(room *lksdk.Room, filename string, fps float64) error {
 	return err
 }
 
-func publishSocket(room *lksdk.Room, addr string, fps float64) error {
-	// Dial Unix socket
-	sock, err := net.Dial("unix", addr)
-	if err != nil {
-		return err
+func parseSocketFromName(name string) (string, string, string, error) {
```
davidzhao, line 181: would be great to have a test case for this.

davidzhao (commented):
this format makes sense! Thanks for your contribution.

Two requests:
* could you add test cases to ensure format is being parsed correctly
* mind updating the README file as well?

bsirang:
@davidzhao I updated the README.

Can you point me to the pattern for writing test cases in this project?

davidzhao:
Hi @bsirang , it could be something [like this](https://github.com/livekit/livekit/blob/master/pkg/service/roomallocator_test.go#L18), where it ensures that different inputs are producing the expected outputs.

bsirang:
@davidzhao thanks, tests added.

davidzhao (approved):
looks good!

PR #68, opened 2022-07-24
chore: Add docs on GStreamer integration

dwirya:
Currently, the support for Unix sockets in GStreamer is via `shmsink`, but I had problems replaying the stream. The solution is to use Unix pipe composed of GStreamer and FFmpeg. GStreamer would use `fdsink` to write to its stdout, and FFmpeg takes that input and forwards it to the socket. Until `shmsink` works out of the box with `ffplay`, I think this solution is the simplest workaround.


dwirya:
Closed in favour of TCP

PR #101, opened 2022-10-02
Support audio sample tracks

davidzhao:
Publishes real test audio instead of random bytes. Helps in assisting
validation of client behavior with audio handling.

This PR also introduces a breaking change for audio bitrate. We can no
longer match audio publishing bitrate by using real samples.


davidzhao:
CC @lukasIO

boks1971 (approved):

davidzhao:
we'll update it when the next CLI version is released.

PR #138, opened 2023-01-27
Match max subscribed participants to grid size, limit to 24

davidzhao:
It does not make sense to subscribe to everything in the room.


boks1971 (approved):

arrase:
Maybe not for you but has a lot of sense for us!! Now we can't simulate our load cases , maybe we misunderstood something? the "--subscribers" parameter still there but i can't have more than 25 subscribers

davidzhao:
Can you share your use case please?

PR #147, opened 2023-02-20
Fix stale room name in load test

msamoylov:
Previously, a generated random room name hasn't been printed. This small change fixes that.

**Before**
```
$ livekit-cli load-test --audio-publishers 32 --subscribers 32 [skipped connection params]
Starting load test with 32 audio publishers, 32 subscribers, room: 
publishing audio track - domji_pub_0
```

**After**
```
$ livekit-cli load-test --audio-publishers 32 --subscribers 32 [skipped connection params]
Starting load test with 32 audio publishers, 32 subscribers, room: testroom405
publishing audio track - sqksj_pub_0
```


davidzhao (approved):
looks good!

PR #148, opened 2023-02-21
Refactor load test for better use of context

msamoylov:
> Contexts should not be stored inside a struct type, but instead passed to each function that needs it.

https://go.dev/blog/context-and-structs

### DIFF
--- a/cmd/livekit-cli/loadtest.go
+++ b/cmd/livekit-cli/loadtest.go
@@ -80,18 +80,18 @@ var LoadTestCommands = []*cli.Command{
 	},
 }
 
-func loadTest(c *cli.Context) error {
-	pc, err := loadProjectDetails(c)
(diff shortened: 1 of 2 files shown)


cmd/livekit-cli/loadtest.go
```diff
@@ -80,56 +80,54 @@ var LoadTestCommands = []*cli.Command{
 	},
 }
 
-func loadTest(c *cli.Context) error {
-	pc, err := loadProjectDetails(c)
+func loadTest(cCtx *cli.Context) error {
```
msamoylov, line 83: Sticking to the [examples](https://cli.urfave.org/v2/examples/full-api-example/) from `urfave/cli`.

davidzhao (approved):
lgtm

PR #160, opened 2023-04-01
v1.2.4

biglittlebigben:
## Changelog

- Support new Ingress protocol
- list-egress --active to display only active Egresses (#162)
- Allowing listing only a single Egress by ID (#151)
- Fix stale room name in load test (#147)
- Added the ability to create recorder tokens (#145)


davidzhao (approved):
nice!

can you also include this other change? would be quite useful: https://github.com/livekit/livekit-cli/pull/162

PR #207, opened 2023-07-27
Add license headers

davidzhao:


frostbyte73 (approved):

cacheonly (approved):
I'm trying to think of reasons why the LICENSE file doesn't suffice... GPTs? we also need to update the copyright date every year

davidzhao:
> I'm trying to think of reasons why the LICENSE file doesn't suffice... GPTs? we also need to update the copyright date every year

We do not need to update every year. new files will carry updated year.

From Apache-2

```
You must retain, in the Source form of any Derivative Works that You distribute, all copyright, patent, trademark, and attribution notices from the Source form of the Work, excluding those notices that do not pertain to any part of the Derivative Works; and

If the Work includes a "NOTICE" text file as part of its distribution, then any Derivative Works that You distribute must include a readable copy of the attribution notices contained within such NOTICE file, excluding those notices that do not pertain to any part of the Derivative Works, in at least one of the following places: within a NOTICE text file distributed as part of the Derivative Works; within the Source form or documentation, if provided along with the Derivative Works; or, within a display generated by the Derivative Works, if and wherever such third-party notices normally appear. The contents of the NOTICE file are for informational purposes only and do not modify the License. You may add Your own attribution notices within Derivative Works that You distribute, alongside or as an addendum to the NOTICE text from the Work, provided that such additional attribution notices cannot be construed as modifying the License.
```

PR #238, opened 2023-10-19
Improve create-room and join-room

davidzhao:
- updated to Go SDK 1.1.1
- added --exit-after-publish option to join-room
- added --min-playout-delay and --sync-streams to create-room


cmd/livekit-cli/room.go
```diff
@@ -43,19 +43,24 @@ var (
...
+				},
+				&cli.StringFlag{
+					Name:  "track-egress-file",
+					Usage: "AutoTrackEgress json file (see examples/auto-track-egress.json)",
 				},
 				&cli.UintFlag{
-					Name:     "min-playout-delay",
-					Usage:    "minimum playout delay",
-					Required: false,
+					Name:  "min-playout-delay",
```
boks1971, line 58: should we add a `max-playout-delay` support as well given that it is supported now?

boks1971 (approved):

cmd/livekit-cli/room.go
davidzhao: perhaps as a separate PR.. wanted to keep this one tight.

PR #254, opened 2023-11-30
Improved list-egress and stop-egress to support batching

davidzhao:
Also providing more information in list-egress table

### DIFF
--- a/cmd/livekit-cli/egress.go
+++ b/cmd/livekit-cli/egress.go
@@ -427,15 +427,20 @@ func updateStream(c *cli.Context) error {
 
 func stopEgress(c *cli.Context) error {
 	ids := c.StringSlice("id")
+	var errors []error
 	for _, id := range ids {
 		_, err := egressClient.StopEgress(context.Background(), &livekit.StopEgressRequest{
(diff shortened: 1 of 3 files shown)


cmd/livekit-cli/egress.go
```diff
@@ -385,14 +426,17 @@ func updateStream(c *cli.Context) error {
 }
 
 func stopEgress(c *cli.Context) error {
-	info, err := egressClient.StopEgress(context.Background(), &livekit.StopEgressRequest{
-		EgressId: c.String("id"),
-	})
-	if err != nil {
-		return err
```
biglittlebigben, line 392: What is the consequence of not returning an error in stopEgress. Does returning an error cause the whole cli to exit with an error code? If so, would changing this behavior potentially break some scripts?

biglittlebigben (approved):

cmd/livekit-cli/egress.go
davidzhao: good point.. I guess we should keep the existing behavior if only a single id is passed in.

biglittlebigben (approved):

PR #260, opened 2023-12-04
Update git command in README

jure:
This is a super small docs PR: When running the current git command, the result is `fatal: repository 'github.com/livekit/livekit-cli' does not exist`. I've followed the example from https://github.com/livekit/livekit/blob/master/README.md#building-from-source and added the `https://` prefix.


davidzhao (approved):
nice catch!

PR #266, opened 2023-12-13
chore: improve dockerfile

chenrui333:
- remove `GO111MODULE` as it is default for go build now
- pin the alpine tag to the latest release `3.19`

cc @Sean-Der


davidzhao (approved):
thanks for the updates

PR #267, opened 2023-12-13
goreleaser: improve binary artifact size

chenrui333:
update ldflags to include `-s -w` to strip the symbol table and DWARF debug information.

cc @Sean-Der


davidzhao (approved):
This helps, thanks!

PR #306, opened 2024-04-01
refine rate limiter to be more accurate

lherman-cs:
The current implementation is not quite representing "NumPerSecond" as it allows the following sequence:

```
Burst -> sleep 1s -> Burst -> ...
```

During the "Burst" time, the load rate can exceed "NumPerSecond" as long as the join can be completed under the "NumPerSecond" interval.

`rate.NewLimiter` can be configured to allow burst as needed. We would need to restructure the code and allow a new parameter to define the burst.


boks1971 (approved):
lgtm!

davidzhao (approved):
hah awesome! I didn't know this lib existed.

PR #340, opened 2024-06-17
Support new SIP Trunk API. Move to subcommands.

dennwc:
Support new SIP Inbound/Outbound Trunk API and move all SIP commands into a separate cli tree:

- `list-sip-trunk` -> `sip trunk list`
- `list-sip-dispatch-rule` -> `sip dispatch list`
- `create-sip-participant` -> `sip participant create`

Old commands continue to work, they are only hidden from the CLI help.

I also renamed the columns for legacy trunks list command. Now they match names in new API (`OutboundNumber` -> `Number`, `InboundNumbers` -> `AllowedNumbers`). 

Requires https://github.com/livekit/protocol/pull/738

### DIFF
--- a/cmd/livekit-cli/sip.go
+++ b/cmd/livekit-cli/sip.go
@@ -17,20 +17,23 @@ package main
 import (
 	"context"
 	"fmt"
-	"os"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/livekit/protocol/livekit"
 	lksdk "github.com/livekit/server-sdk-go/v2"
-	"github.com/olekukonko/tablewriter"
 	"github.com/urfave/cli/v2"
 )
 
 //lint:file-ignore SA1019 we still support older APIs for compatibility
 
-const sipCategory = "SIP"
+const (
+	sipCategory            = "SIP"
+	sipTrunkCategory       = "Trunks"
+	sipDispatchCategory    = "Dispatch Rules"
+	sipParticipantCategory = "Participants"
+)
 
 var (
 	SIPCommands = []*cli.Command{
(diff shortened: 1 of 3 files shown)


frostbyte73 (approved):

dennwc:
This should wait for prod deployment. It hides old CLI commands, while new ones won't work without updated server.

dennwc:
Regarding sub-commands, it would be also nice to move shared options like URL/key/secret to the root command:
```
livekit-cli --api-key <key> --api-secret <secret> some sub command
```
instead of
```
livekit-cli some sub command  --api-key <key> --api-secret <secret>
```

rektdeckard:
@davidzhao @dennwc  looks like what what we had in mind, and consistent with prior art in the `projects` subcommand. Great shout on moving the global parameters too! Other main thing to think about IMO is making the key argument a positional argument, for example

```sh
livekit-cli room join "my-first-room"
```

instead of

```sh
livekit-cli join-room --room "my-first-room"
```

Since it is the main target of the command. I don't know if the `--request` flag qualifies here or not. WDYT?

rektdeckard (approved):

dennwc:
Good idea! Will make main arg positional then for new SIP commands :+1: 

Regarding flags, that should probably be in a separate PR, because existing commands will need to be updated too.

dennwc:
@rektdeckard Updated new SIP commands to accept main args (`id` or `request`).

Moving flags to the top will require an update to `github.com/urfave/cli/v3`, which added `Persistent` to the flag definitions. Could be a separate PR.

cmd/livekit-cli/sip.go
```diff
@@ -283,62 +423,74 @@ func listSipDispatchRule(c *cli.Context) error {
...
 	if err != nil {
 		return err
 	}
-
-	printSIPParticipantInfo(info)
-	return nil
+	return createAndPrintLegacy(c, func(ctx context.Context, req *livekit.CreateSIPParticipantRequest) (*livekit.SIPParticipantInfo, error) {
+		// CreateSIPParticipant will wait for LiveKit Participant to be created and that can take some time.
+		// Default deadline is too short, thus, we must set a higher deadline for it.
+		ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
rektdeckard, line 489: Probably outside my wheelhouse, but when docs are updated it may be good to include constants like this if it is invariant, or expose it as a parameter if not.

rektdeckard:
Nice! Agree, separate PR for that. `v2` seems to be in maintenance mode, so probably a good thing regardless?

rektdeckard (approved):

cmd/livekit-cli/sip.go
dennwc: You are right. I can actually bake it into the SDK, so not configuration will be needed.

PR #343, opened 2024-06-21
feat(cli): restructure subcommands

rektdeckard:
- Renames the CLI binary and associated commands from `livekit-cli` to `lk` [LK-3475](https://linear.app/livekit/issue/LK-3475/rename-livekit-cli-to-lk)
  - [x] Build from source
  - [ ] Linux install script (`curl -sSL https://get.livekit.io/cli | bash`) must wait for GitHub release
  - [ ] macOS Homebrew formula (`brew install livekit-cli`)
    - [x] Install procedure updated ([rektdeckard/homebrew-core](https://github.com/rektdeckard/homebrew-core/tree/bump-livekit-cli-2.0.0)]
    - [ ] Regenerate formula hashes once GitHub release is live with `brew bump-formula-pr livekit-cli`
  - [x] Windows and other precompiled binaries (CI)
  - [x] Symlinks `livekit-cli` -> `lk` on install, and prints a deprecation warning message if the binary detects `$0` is not `lk`
- Upgrades to `urfave/cli/v3` [LK-3479](https://linear.app/livekit/issue/LK-3479/upgrade-to-urfavecliv3)
- Restructures many of the commands and subcommands to be more discoverable [LK-3477](https://linear.app/livekit/issue/LK-3477/reorganize-cli-subcommands)
- Improves a lot of the usage text

Do not merge before https://github.com/livekit/web/pull/238


install-cli.sh
```diff
@@ -21,6 +21,7 @@ set -o errexit
 set -o pipefail
 
 REPO="livekit-cli"
+BIN_NAME="lk"
```
davidzhao, line 24: let's update the script last. the main reason is that whatever is merged in main is going to be pulling the last release from GH. so if we did it now it'll break previous installs

davidzhao (commented):
given this is a breaking change, I think we should be careful when pushing it. I would advocate for:

- marking the version as 2.0, to signify the new name + organization
- including a `livekit-cli` binary, that when ran, it just lets users know about the change
- update all of our docs/repos for references to livekit-cli
- ensuring that Homebrew formula is updated to work with the new CLI

install-cli.sh
rektdeckard: Good idea -- I'll split off the install script changes into a separate PR. Can land that and docs site changes around the same time.

Like the idea of a dummy `livekit-cli` binary too!

rektdeckard:
Docs changes PR (WIP) is https://github.com/livekit/web/pull/238

cmd/lk/ingress.go
```diff
@@ -32,6 +32,73 @@ const ingressCategory = "Ingress"
...
+			Name:     "ingress",
+			Usage:    "Import outside media sources into a LiveKit room",
+			Category: "I/O",
+			Commands: []*cli.Command{
+				{
+					Name:   "create",
+					Usage:  "Create an ingress",
+					Before: createIngressClient,
+					Action: createIngress,
+					Flags: withDefaultFlags(
```
dennwc, line 44: `withDefaultFlags` is no longer necessary for `urfave/cli/v3`. We can now move common flag to the root command and set `Persistent: true` on these flags. They should then be available on all subcommands.

cmd/livekit-cli/main.go
```diff
@@ -1,97 +1,11 @@
...
-	if outPath != "" {
-		if err := os.WriteFile(outPath, []byte(fishScript), 0o644); err != nil {
-			return err
-		}
-	} else {
-		fmt.Println(fishScript)
-	}
-
-	return nil
+	fmt.Fprintln(os.Stderr, `The "livekit-cli" binary has been renamed to "lk". Try again with the new name!`)
```
dennwc, line 9: I wonder, do we really have to do this? Can we symlink instead? CLI could also check its own binary name (aka `os.Args[0]`) and print a migration warning.

Since we will hide all CLI commands and they will continue to work anyway, there's no real breaking change here. Why are we specifically breaking the CLI under the old name?

cmd/lk/main.go
```diff
@@ -0,0 +1,100 @@
...
+				logConfig.Level = "debug"
+			}
+			logger.InitFromConfig(logConfig, "livekit-cli")
+			lksdk.SetLogger(logger.GetLogger())
+
+			return nil
+		},
+	}
+
+	app.Commands = append(app.Commands, TokenCommands...)
```
dennwc, line 70: I've seen a pretty convenient CLI pattern in other projects:

```go
// main.go
var App = &cli.Command{
		Name:                  "lk",
		// ...
}

func main() {
	if err := app.Run(context.Background(), os.Args); err != nil {
		fmt.Println(err)
	}
}
```

```go
// some-sub-command.go

func init() {
	cmd := &cli.Command{
		Name:                  "some-sub-cmd",
		// ...
	}
	App.Commands = append(App.Commands, cmd)
}
```

That way each CLI sub-command file can register its own list of commands without the need to pile all registrations in main.

cmd/lk/main.go
```diff
@@ -0,0 +1,100 @@
...
+	app.Commands = append(app.Commands, RoomCommands...)
+	app.Commands = append(app.Commands, JoinCommands...)
+	app.Commands = append(app.Commands, EgressCommands...)
+	app.Commands = append(app.Commands, IngressCommands...)
+	app.Commands = append(app.Commands, LoadTestCommands...)
+	app.Commands = append(app.Commands, ProjectCommands...)
+	app.Commands = append(app.Commands, SIPCommands...)
+
+	if err := app.Run(context.Background(), os.Args); err != nil {
+		fmt.Println(err)
```
dennwc, line 80:
```suggestion
		fmt.Fprintln(os.Stderr, err)
		os.Exit(1)
```

cmd/lk/main.go
```diff
@@ -0,0 +1,100 @@
...
+			},
+		},
+		Before: func(ctx context.Context, cmd *cli.Command) error {
+			logConfig := &logger.Config{
+				Level: "info",
+			}
+			if cmd.Bool("verbose") {
+				logConfig.Level = "debug"
+			}
+			logger.InitFromConfig(logConfig, "livekit-cli")
```
dennwc, line 63: Will it still call itself `livekit-cli` in logs?

cmd/lk/loadtest.go
```diff
@@ -98,18 +98,18 @@ var LoadTestCommands = []*cli.Command{
...
 	}
 
-	if !cCtx.Bool("verbose") {
+	if !cmd.Bool("verbose") {
 		lksdk.SetLogger(logger.LogRLogger(logr.Discard()))
 	}
 	_ = raiseULimit()
 
-	ctx, cancel := context.WithCancel(cCtx.Context)
+	ctx, cancel := context.WithCancel(ctx)
```
dennwc, line 112: This should probably use [NotifyContext](https://pkg.go.dev/os/signal@go1.22.4#NotifyContext).

cmd/lk/utils.go
```diff
@@ -96,6 +98,34 @@ func withDefaultClientOpts(c *config.ProjectConfig) []twirp.ClientOption {
 	return opts
 }
 
+func extractArg(c *cli.Command) (*string, error) {
+	if !c.Args().Present() {
+		return nil, errors.New("no argument provided")
+	}
+	return &c.Args().Slice()[0], nil
```
dennwc, line 105: Isn't there an `c.Args().First()` helper?

cmd/lk/ingress.go
```diff
@@ -54,12 +122,13 @@ var (
 			Flags: withDefaultFlags(
 				&cli.StringFlag{
 					Name:     "request",
-					Usage:    "UpdateIngressRequest as json file (see livekit-cli/examples)",
+					Usage:    "UpdateIngressRequest as json file (see cmd/livekit-cli/examples)",
```
dennwc, line 125: `livekit-cli` -> `lk`? Same in other places.

cmd/lk/main.go
```diff
@@ -0,0 +1,100 @@
...
+	app.Commands = append(app.Commands, TokenCommands...)
+	app.Commands = append(app.Commands, RoomCommands...)
+	app.Commands = append(app.Commands, JoinCommands...)
+	app.Commands = append(app.Commands, EgressCommands...)
+	app.Commands = append(app.Commands, IngressCommands...)
+	app.Commands = append(app.Commands, LoadTestCommands...)
+	app.Commands = append(app.Commands, ProjectCommands...)
+	app.Commands = append(app.Commands, SIPCommands...)
+
+	if err := app.Run(context.Background(), os.Args); err != nil {
```
dennwc, line 79: It appears that `Run` doesn't listen to OS signals by default, so worth using `signal.NotifyContext` instead of `context.Background` here.

README.md
```diff
@@ -81,7 +81,7 @@ livekit-cli project set-default <project-name>
 To publish a demo video as a participant's track, use the following.
 
 ```shell
-livekit-cli join-room --room yourroom --identity publisher \
+lk join-room --room yourroom --identity publisher \
```
dennwc, line 84: `lk room join` maybe?

README.md
```diff
@@ -159,17 +159,17 @@ ffplay -i unix:/tmp/myvideo.sock
 
 Recording requires [egress service](https://docs.livekit.io/guides/egress/) to be set up first.
 
-Example request.json files are [located here](https://github.com/livekit/livekit-cli/tree/main/cmd/livekit-cli/examples).
+Example request.json files are [located here](https://github.com/livekit/livekit-cli/tree/main/cmd/lk/examples).
 
 ```shell
 # start room composite (recording of room UI)
-livekit-cli start-room-composite-egress --request request.json
+lk start-room-composite-egress --request request.json
```
dennwc, line 166: `lk egress start room-composite`?

README.md
```diff
@@ -159,17 +159,17 @@ ffplay -i unix:/tmp/myvideo.sock
...
 
 # start track composite (audio + video)
-livekit-cli start-track-composite-egress --request request.json
+lk start-track-composite-egress --request request.json
 
 # start track egress (single audio or video track)
-livekit-cli start-track-egress --request request.json
+lk start-track-egress --request request.json
```
dennwc, line 172: `lk egress start track`?

README.md
```diff
@@ -183,7 +183,7 @@ It'll then open a browser to the template URL, with the correct parameters fille
 Here's an example:
 
 ```shell
-livekit-cli test-egress-template \
+lk test-egress-template \
```
dennwc, line 186: `lk test egress-template` or `lk egress test template`?

README.md
```diff
@@ -25,7 +25,7 @@ This package includes command line utilities that interacts with LiveKit. It all
 ## Mac
 
 ```shell
-brew install livekit-cli
+brew install lk
```
dennwc, line 28: Is it possible to bind short names like this on Brew?

cmd/lk/main.go
rektdeckard: Only issue with that is that order matters. If commands self-register, the ordering in Usage is at the whims of the compiler. I'd rather have control over that, even if it's a little awkward.

README.md
rektdeckard: Nope! We aren't renaming the formula. Readme changes are stale, I'll fix this.

cmd/livekit-cli/main.go
rektdeckard: I think, because we can with a 2.0 release? Yes we support the old commands, but we don't want to actively encourage their use and they will not be maintained.

I agree a symlink and detecting $0 is more elegant than a separate (~1.4MB) binary, but it's also more fiddly. Not that it's hard, but a simple solution is also simple to remove when we feel ready.

cmd/lk/main.go
rektdeckard: Nope, missed that!

cmd/lk/utils.go
rektdeckard: 🤦

README.md
rektdeckard: The way I refactored it, it is now `lk egress start --type track` (when omitted, type defaults to `room-composite`)

rektdeckard: Also I'm new to Go, but we could also consider doing a tagged union approach where `request.json` has a `type` field we can read and know implicitly the kind of struct to unmarshal, and what operation you're doing without having to be redundant.

cmd/livekit-cli/main.go
dennwc: > I think, because we can with a 2.0 release?

Right, if you proposed to remove old CLI _and_ old sub-commands for v2 that would be understandable. But we are keeping old commands still. So why not keep the binary as well (as a symlink at least)?

> but it's also more fiddly

Definitely more fiddly for us, but the users might appreciate it :man_shrugging:

cmd/lk/utils.go
rektdeckard: OK, I believe my thinking was that you can't return `nil` in place of a `string`, so would have to `return "", err` in case of an error -- which felt wrong? IDK I am still learning what idiomatic Go is 😆

cmd/lk/main.go
rektdeckard: Hmm, I though `Background()` used the default behaviors got a context?
https://pkg.go.dev/os/signal#hdr-Default_behavior_of_signals_in_Go_programs

cmd/lk/ingress.go
rektdeckard: Done

cmd/lk/utils.go
dennwc: Returning empty string is totally OK in go :+1:

cmd/lk/loadtest.go
rektdeckard: Actually, is this wrapped context even needed? The ancestor context at the root of the app is now a `NotifyContext`, and it looks like the purpose of this code and the following lines is to do the same thing. It is already going to get cleaned up when the program receives these signals.

dennwc: You are right! It does handle more signals, but I don't think it matters for CLI.

Maybe worth doing `git blame` and ask the author why these signals are handled explicitly here.

rektdeckard: Paging @davidzhao

rektdeckard: I think I see what it was intending to do, but it doesn't really do that currently. In the release build, If you try to Ctrl-C while the load test is being set up, it's ignored:

```
tobiasfried@slate livekit-cli % livekit-cli load-test --duration 10m
Using default project local
Starting load test with 1 video publishers, 1 subscribers, room: testroom467
^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C^C im trying to kill itcould not connect Sub 0: could not dial signal connection
could not connect Pub 0: could not dial signal connection
Finished connecting to room, waiting 10m0s

Sub 0 | Track | Kind | Pkts | Bitrate | Dropped

Summary | Tester | Tracks | Bitrate               | Total Dropped | Error
        | Sub 0  | 0/1    | NaNmbps               |  -            | could not dial signal connection
        | Total  | 0/1    | NaNmbps (NaNmbps avg) |  -            | 1
tobiasfried@slate livekit-cli %
```
vs now

```
tobiasfried@slate livekit-cli % bin/lk load-test --duration 10m
Using default project local
Starting load test with 1 video publishers, 1 subscribers, room: testroom851
^C
tobiasfried@slate livekit-cli %
```

rektdeckard: Looks like the separate context in the original code is there to allow the load tester a chance to print the summary before the program terminates after a SIGINT. It relies on the fact that the root context was *not* hooking these signals, and that the contexts would unwind in reverse.

New code accomplishes the same thing with a single context, so I'm pretty sure its OK?

cmd/livekit-cli/main.go
davidzhao: +1 to including a warning, but don't break immediately

davidzhao (commented):
lg! will defer to @dennwc to approve once README changes are resolved

cmd/lk/loadtest.go
dennwc: Yeah, I was mostly wondering about the additional signals. Apart from that it should be good.

dennwc: Also, I don't think it's required to wait for the summary. We should exit as soon as possible, if user asks us to.

rektdeckard: With utilities like `load-test` and `egress test-template`, the default duration could be a lot longer than the user wants to wait, or they could simply get bored and be happy with a shorter sample of data. It feels like good DX to me to still print this, and it's just a tiny bit of data jockeying and string formatting.

rektdeckard: And because of the way the hook is unregistered, no matter what happens, sending a second SIGINT or whatever will always insta-terminate

rektdeckard: True. I think that's a future refactor that requires moving things around a bit. Suffice it to say that this https://github.com/livekit/livekit-cli/pull/343/files#diff-16e3a831829b9fec5dc26d7fa057e9d0326c598b0be63ff443a7794f271e4705L344-R365 combined with using a root NotifyContext, accomplishes the same thing for now.

cmd/livekit-cli/main.go
rektdeckard: OK, removed the dummy binary and added symlinking. The CLI will now print the following warning if `os.Args[0]` does not end with "lk":

```
tobiasfried@slate bin % livekit-cli --help

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ DEPRECATION NOTICE ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
The `livekit-cli` binary has been renamed to `lk`, and some of the options and
commands have changed. Though legacy commands my continue to work, they have
been hidden from the USAGE notes and may be removed in future releases.
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

NAME:
   lk - CLI client to LiveKit

USAGE:
   lk [global options] [command [command options]] [arguments...]

VERSION:
   2.0.0

DESCRIPTION:
   A suite of command line utilities allowing you to access LiveKit APIs services, interact with rooms in realtime, and perform load testing simulations.

COMMANDS:
   Core:
     token    Create access tokens with granular capabilities
     room     Create or delete rooms and manage existing room properties
     project  Add or remove projects and view existing project properties
   I/O:
     egress   Record or stream media from LiveKit to elsewhere
     ingress  Import outside media sources into a LiveKit room
     sip      Manage SIP Trunks, Dispatch Rules, and Participants
   Simulate:
     load-test  Run load tests against LiveKit with simulated publishers & subscribers

GLOBAL OPTIONS:
   --url URL            URL to LiveKit instance (default: "http://localhost:7880") [$LIVEKIT_URL]
   --api-key KEY        Your KEY [$LIVEKIT_API_KEY]
   --api-secret SECRET  Your SECRET [$LIVEKIT_API_SECRET]
   --project NAME       NAME of a configured project
   --curl               Print curl commands for API actions (default: false)
   --verbose            (default: false)
   --help, -h           show help (default: false)
   --version, -v        print the version (default: false)
```

rektdeckard:
@dennwc @davidzhao  Ready for final review, I think. Couple things to note:

The Linux install script and the brew formula both rely on GitHub release artifacts, so they will come in a separate PR here and [against homebrew](https://github.com/rektdeckard/homebrew-core/tree/bump-livekit-cli-2.0.0), respectively, after this is merged and released. 

Since our only means of distribution for Windows is pre-built binaries, I haven't done anything besides change the name of the binary we build. Those users will probably figure it out for themselves when they upgrade (though we might want to consider packaging for `choco` and/or `winget`).

And of course, all of this should line up with https://github.com/livekit/web/pull/238

davidzhao (approved):
lgtm!

PR #353, opened 2024-07-06
Use participant identity when sending data

davidzhao:


rektdeckard (approved):
Just to be more consistent with the other command APIs, I guess? Or is `DestinationIdentities` a new thing?

davidzhao:
> Just to be more consistent with the other command APIs, I guess? Or is `DestinationIdentities` a new thing?

Yeah it's sort of new. In V2 of our client sdks, we've moved away from expose session IDs in public APIs: https://docs.livekit.io/recipes/migrate-from-v1/#Updated-publishData-API

PR #356, opened 2024-07-06
chore(cli): bump version

rektdeckard:


req.json
davidzhao, line 1: unintentional?

davidzhao (approved):

req.json
rektdeckard: yes, gone now! testing that trailing comma thing

PR #360, opened 2024-07-10
fix(room join): room name is a positional argument

schelmo:


schelmo:
sorry i don't want sign a CLA. see #361

PR #370, opened 2024-07-11
add replay commands

frostbyte73:


cmd/lk/replay.go
```diff
@@ -0,0 +1,223 @@
...
+					Action: listReplays,
+				},
+				{
+					Name:   "load",
+					Before: createReplayClient,
+					Action: loadReplay,
+					Flags: []cli.Flag{
+						&cli.StringFlag{
+							Name:     "id",
+							Usage:    "replay ID",
```
rektdeckard, line 39: For all named flags, you can add backticks around a capitalized string to automatically use it as the arg name in Usage. We're also using Sentence case for the usage strings now, just for legibility.

```suggestion
							Usage:    "Replay `ID`",
```

rektdeckard (approved):
Looks good aside from usage string nits!

rektdeckard:
The other flags that take arguments could use backtick treatment too.
```
Usage: "Playback room `NAME`"
```
 gets rendered in the help as `--room NAME` instead of `--room value`.

rektdeckard:
Obviously this is all hidden now, so not pressing.

PR #371, opened 2024-07-13
feat(cmd): auth subcommand

rektdeckard:


cmd/lk/cloud.go
```diff
@@ -0,0 +1,319 @@
...
+	disconnect   bool
+	timeout      int64
+	interval     int64
+	serverURL    string
+	dashboardURL string
+	authClient   AuthClient
+	AuthCommands = []*cli.Command{
+		{
+			Name:     "cloud",
+			Usage:    "Manage hosted services and authentication",
```
davidzhao, line 51:
```suggestion
			Usage:    "Interacting with LiveKit Cloud",
```

cmd/lk/cloud.go
```diff
@@ -0,0 +1,319 @@
...
+	err = json.NewDecoder(resp.Body).Decode(&ak)
+	if err != nil {
+		return nil, err
+	}
+
+	return ak, nil
+}
+
+func (a *AuthClient) Deauthenticate() error {
+	// TODO: revoke any session token
```
davidzhao, line 182: intentional?

cmd/lk/cloud.go
```diff
@@ -0,0 +1,319 @@
...
+		return err
+	}
+
+	authURL, err := generateConfirmURL(token.Token)
+	if err != nil {
+		return err
+	}
+
+	// poll for keys
+	fmt.Printf("Please confirm access by visiting:\n\n   %s\n\n", authURL.String())
```
davidzhao, line 239: it's also possible to launch this page too: [browser.OpenURL](https://github.com/livekit/livekit-cli/blob/main/cmd/lk/egress.go#L756)

pkg/config/config.go
```diff
@@ -37,6 +37,19 @@ type ProjectConfig struct {
 	APISecret string `yaml:"api_secret"`
 }
 
+type Project struct {
+	Subdomain string
+}
+
+type AccessKey struct {
```
davidzhao, line 44: what do these do?

cmd/lk/cloud.go
rektdeckard: 🆒 I didn't see that. Is it well-supported? Guessing it uses `xdg-open` on linux, `open` on mac, `start` on windows.

rektdeckard: Probably appropriate to use both methods, in case user has some popup blocking settings or smth.

pkg/config/config.go
rektdeckard: These are temporary types we deserialize into before we write project config to our config file. We don't have access to `backend-common` here due to OSS, so basically just recreating a minimal struct that has only the parts we need.

cmd/lk/cloud.go
davidzhao: yah, always good to show the URL, but this has worked pretty well thus far.

davidzhao (approved):
lg!

cmd/lk/cloud.go
rektdeckard: Was waiting on https://github.com/livekit/cloud-api-server/pull/747

PR #385, opened 2024-08-13
fix(lk): correct cfg file permission check and bin name

rektdeckard:
- Fixes erroneous file permission check on config directory
- Explicit build targets to make it simpler to build in Windows
- Deprecation check is now aware of Windows executable name

https://linear.app/livekit/issue/CLT-560/cli-error-config-file-cusersgrafhlivekitcli-configyaml-should-be-0600

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -83,7 +83,9 @@ func LoadOrCreate() (*CLIConfig, error) {
 		return c, nil
 	} else if err != nil {
 		return nil, err
-	} else if s.Mode().Perm()&0600 != 0600 {
+		// because this file contains private keys, ensure
+		// that only the owner has permission to access it
+	} else if s.Mode().Perm()&0077 != 0 {
 		return nil, fmt.Errorf("config file %s should be 0600", configPath)
 	}
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -83,7 +83,7 @@ func LoadOrCreate() (*CLIConfig, error) {
 		return c, nil
 	} else if err != nil {
 		return nil, err
-	} else if s.Mode().Perm()&0077 != 0 {
+	} else if s.Mode().Perm()&0600 != 0600 {
 		return nil, fmt.Errorf("config file %s should be 0600", configPath)
 	}
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -83,10 +83,10 @@ func LoadOrCreate() (*CLIConfig, error) {
 		return c, nil
 	} else if err != nil {
 		return nil, err
-		// because this file contains private keys, ensure
-		// that only the owner has permission to access it
 	} else if s.Mode().Perm()&0077 != 0 {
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -83,7 +83,7 @@ func LoadOrCreate() (*CLIConfig, error) {
 		return c, nil
 	} else if err != nil {
 		return nil, err
-	} else if s.Mode().Perm()&0077 != 0 {
+	} else if s.Mode().Perm()&0600 != 0600 {
```
davidzhao, line 86: i'm not sure if this is the right check.. if I set my file to 0777, 0777 & 0600 is still 0600?

rektdeckard: The correct check should be to make sure that the file has _at least_ 0600 (owner can read and write), not _exactly_ 0600, no? I set up my Windows dev machine by copying my config file over to it rather than having it generated, and while I definitely had read and write permissions on the file, it was not exactly 0600. This caused an error and was unexpected to me

mike-r-mclaughlin: this looks right to me if we only care to check that the user has allowed us to read/write the file and we don't want to warn the user about the potential security concern of opening the file permissions too wide.  i.e., we are potentially allowing the file to be completely open (0777) to anyone on the same system.

the main reasons to force it to be 0600 are a). confirm we can read and write the file and b) stop the user from potentially exposing their API key/secret to others on the system if the permissions for group/other are non-zero.  

alternatively, maybe we just print a warning message if it's not 0600 (and probably is safe to ignore 0700) rather than throw an error?

davidzhao: for security purposes, the idea is to ensure that group and other could not read or write. that would match SSH's permission enforcement on the private key: https://superuser.com/a/215506

rektdeckard: The problem I was trying to address here is that on Windows, the unix-equivalent file permissions for files you create or copy into `%USERPROFILE%` seem to be 0750, not 0700. I'm happy to revert this change if the friction is worth the security.

davidzhao: trying to understand the friction here.. I agree we should avoid it if possible.

we create the file [here](https://github.com/livekit/livekit-cli/blob/main/pkg/config/config.go#L145) with `0600`, are you saying it somehow changed to `0750`? 

how did that work?

rektdeckard: Both me (and the user in [this linear ticket](https://linear.app/livekit/issue/CLT-560/cli-error-config-file-cusersgrafhlivekitcli-configyaml-should-be-0600), presumably) copied our config file from another machine onto their windows machine, rather than having it generated for us. The default file permissions were inherited from the parent folder (0750) and thus were too lax.

My thinking was we could relax the permission check to allow the user to make their own decisions with file permissions, but I guess I'm convinced now we should just be strict.

mike-r-mclaughlin: couldn't we just make it a warning message but allow them to continue?

rektdeckard: i like it

davidzhao (approved):
lg

PR #387, opened 2024-08-13
feat(lk): initial pass at app create command

rektdeckard:


Ocupe (approved):
Oops, just saw the latest change. Haven't reviewed the full PR yet.

cmd/lk/app.go
```diff
@@ -0,0 +1,379 @@
...
+		"LIVEKIT_API_SECRET": project.APISecret,
+		"LIVEKIT_URL":        project.URL,
+		"LIVEKIT_SANDBOX":    "TODO",
+	}
+
+	prompt := func(key, oldValue string) (string, error) {
+		var newValue string
+		if err := huh.NewInput().
+			Title("Enter " + key + "?").
+			Placeholder(oldValue).
```
nbsp, line 304: re: demo, maybe this should be a password field? it's being pasted anyway

```diff
			Placeholder(oldValue).
+ 			EchoMode(textinput.EchoPassword).
```

rektdeckard: Solid idea. I leaked my keys 5 ways in that demo 😆

rektdeckard: At the same time, not all of these are passwords/secrets. Our livestream example app actually uses an environment variable that stores a whole yaml document with a UI configuration. That could be quite annoying. Maybe we just trust the devs to have some opsec?

nbsp: hrm, yeah. maybe a heuristic that searches for `SECRET` or `KEY` or `PASSWORD`?

PR #393, opened 2024-08-20
Send and receive DTMF in join command.

dennwc:
Switch to the new `OnDataPacket` callback instead of the legacy `OnDataReceived`. Add flags for publishing DTMF and user data to `lk room join`. Legacy `lk join-room` still uses old callback and doesn't support new flags.

### DIFF
--- a/cmd/lk/room.go
+++ b/cmd/lk/room.go
@@ -152,6 +152,14 @@ var (
 								"Can be used multiple times to publish multiple files. " +
 								"Can publish from Unix or TCP socket using the format '<codec>://<socket_name>' or '<codec>://<host:address>' respectively. Valid codecs are \"h264\", \"vp8\", \"opus\"",
 						},
+						&cli.StringFlag{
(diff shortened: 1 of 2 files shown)


cmd/lk/room.go
```diff
@@ -820,11 +835,12 @@ func joinRoom(ctx context.Context, cmd *cli.Command) error {
 		}
 	}
 
-	if cmd.StringSlice("publish") != nil {
+	exitAfterPublish := cmd.Bool("exit-after-publish")
+	if publish := cmd.StringSlice("publish"); publish != nil {
```
rektdeckard, line 839: Just confirming this returns nil and not an empty slice if flag is omitted?

rektdeckard (approved):
Looks good to me

cmd/lk/room.go
dennwc: I assume so, if it was working before :man_shrugging:

PR #396, opened 2024-08-21
chore(auth): cleanup copy, dedupe project names, and enable feature

rektdeckard:
- Enable `lk cloud auth` for public use
- Use project names instead of device names for stored credentials
- Detect and prompt for duplicate project names
- Updated usage copy for `lk cloud auth`

### DIFF
--- a/cmd/lk/cloud.go
+++ b/cmd/lk/cloud.go
@@ -36,6 +36,7 @@ type ClaimAccessKeyResponse struct {
 	Key         string
 	Secret      string
 	ProjectId   string
+	ProjectName string
 	OwnerId     string
 	Description string
 	URL         string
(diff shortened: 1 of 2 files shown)


cmd/lk/cloud.go
```diff
@@ -322,16 +322,34 @@ func tryAuthIfNeeded(ctx context.Context, cmd *cli.Command) error {
 		return err
 	}
 
+	// make sure name is unique
+	var name = ak.ProjectName
+	if cliConfig.ProjectExists(name) {
+		if err := huh.NewInput().
```
davidzhao, line 328: for some reasons.. every time I see this package, I can't help but to think of @cacheonly and Noah's joke

pkg/config/config.go
```diff
@@ -103,6 +103,15 @@ func LoadOrCreate() (*CLIConfig, error) {
 	return c, nil
 }
 
+func (c *CLIConfig) ProjectExists(name string) bool {
+	for _, p := range c.Projects {
+		if p.Name == name {
```
davidzhao, line 108: can we do a case insensitive comparison? just in case.

davidzhao (approved):
lg

cmd/lk/cloud.go
rektdeckard: SAME https://github.com/livekit/livekit-cli/pull/365#issue-2401952169

pkg/config/config.go
rektdeckard: yes, good call!

cmd/lk/cloud.go
rektdeckard: What was Noah's joke though?

PR #399, opened 2024-08-27
Fix flags getting erroneously hidden or optional

davidliu:
The optional/hidden functions modify the original flag passed in rather than create their own copy. This ends up displaying the wrong options in the help menu for a command. For example, `lk room send-data` requires the `--room` flag, but running help on the command gives this:

```
NAME:
   lk room send-data - Send arbitrary JSON data to client

USAGE:
   lk room send-data [OPTIONS] DATA

OPTIONS:
   --topic TOPIC                          TOPIC of the message
   --identity value [ --identity value ]  One or more participant identities to send the message to. When empty, broadcasts to the entire room
   --help, -h                             show help (default: false)
```

It also allows running the command itself, not raising that the `room` flag wasn't set.

### DIFF
--- a/cmd/lk/utils.go
+++ b/cmd/lk/utils.go
@@ -86,15 +86,15 @@ var (
 )
 
 func optional[T any, C any, VC cli.ValueCreator[T, C]](flag *cli.FlagBase[T, C, VC]) *cli.FlagBase[T, C, VC] {
-	newFlag := flag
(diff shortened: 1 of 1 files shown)


rektdeckard:
Thanks for spotting!

rektdeckard (approved):

PR #410, opened 2024-09-12
chore(cmd): cleanup for sbx internal alpha

rektdeckard:
- Minor cleanup around sandbox/bootstrap commands
- Bump version to `2.1.1`. Though you could consider this a `2.2.0-alpha.1`, it's easier to release this way for homebrew and winget.

### DIFF
--- a/cmd/lk/app.go
+++ b/cmd/lk/app.go
@@ -326,6 +326,7 @@ func cloneTemplate(_ context.Context, cmd *cli.Command, url, appName string) err
 			}
 			os.RemoveAll(path.Join(appName, ".git"))
 		}).
+		Style(theme.Focused.Title).
 		Run(); err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


cmd/lk/app.go
```diff
@@ -87,18 +87,12 @@ var (
 						},
 						&cli.StringFlag{
 							// Use "http://cloud-api.livekit.run" in local dev
+							// Use "https://cloud-api-server-staging.ochicago1a.staging.livekit.app" in staging
```
davidzhao, line 90: any way to keep these out of the OSS repo?

davidzhao (commented):
just a couple of nits

cmd/lk/app.go
rektdeckard: We can remove it, just there as a courtesy 👍

davidzhao (approved):

PR #417, opened 2024-09-16
feat(cmd): add next steps hint after template install

rektdeckard:
- Reverts to using default terminal color palette, since branded colors do not always play nice with light terminal themes
- Adds hint text to guide users to start agent after install

Screenshot 2024-09-16 at 5 01 31 PM

### DIFF
--- a/cmd/lk/app.go
+++ b/cmd/lk/app.go
@@ -88,9 +88,16 @@ var (
 				{
 					Name:      "install",
 					Usage:     "Execute installation defined in " + bootstrap.TaskFile,
-					ArgsUsage: "`DIR` location or the project directory (default: current directory)",
+					ArgsUsage: "[DIR] location of the project directory (default: current directory)",
 					Before:    requireProject,
 					Action:    installTemplate,
+				},
+				{
+					Hidden:    true,
+					Name:      "run",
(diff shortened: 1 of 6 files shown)


cmd/lk/app.go
```diff
@@ -88,10 +88,17 @@ var (
...
 					Before:    requireProject,
 					Action:    installTemplate,
 				},
+				{
+					Hidden:    true,
+					Name:      "run",
+					Usage:     "Execute a task defined in " + bootstrap.TaskFile,
+					ArgsUsage: "[TASK] to run in the project's taskfile.yaml",
+					Action:    runTask,
+				},
```
rektdeckard, line 101: @davidzhao hiding this command for now. my other idea was to have a `post_instantiate` task or smth that runs after we clone, degit, and inflate the environment. All it would do is print the steps to install and get started:

```bash
tobiasfried@slate /tmp > lk app sandbox --id 5lfwjtkd6cy
Using default project [Atrium]
Cloning template...
Instantiating environment...
Done! To setup and run your sandbox:

   cd /tmp/foo
   python3 -m venv venv
   python3 -m pip install -r requirements.txt
   python3 agent.py start

tobiasfried@slate /tmp >
```

PR #418, opened 2024-09-17
feat(cmd): use written install instructions instead of automatic

rektdeckard:
Instead of taking the place of tools devs are already familiar with, just print out the steps for devs to run themselves.

```sh
tobiasfried@slate /tmp % lk app sandbox --id 5lfwjtkd6cy
Using default project [Atrium]
Cloning template...
Instantiating environment...
Running task post_create_sandbox...
To setup and run the agent:

    cd /tmp/optimized-instance/agent
    python3 -m venv venv
    source venv/bin/activate
    python3 -m pip install -r requirements.txt
    python3 agent.py start

tobiasfried@slate /tmp %
```

### DIFF
--- a/cmd/lk/app.go
+++ b/cmd/lk/app.go
@@ -21,7 +21,6 @@ import (
 	"os"
 	"os/exec"
 	"path"
-	"path/filepath"
 	"regexp"
 
 	"github.com/charmbracelet/huh"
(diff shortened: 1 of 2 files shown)


rektdeckard:
@davidzhao yep, every app customizes the (optional) text that appears on initializing a sandbox, and another on bootstrap.

PR #425, opened 2024-09-19
Implement TransferParticipant in SIP

biglittlebigben:


cmd/lk/sip.go
```diff
@@ -128,6 +129,12 @@ var (
 							Action:    createSIPParticipant,
 							ArgsUsage: RequestDesc[livekit.CreateSIPParticipantRequest](),
 						},
+						{
+							Name:      "transfer",
+							Usage:     "Transfer a SIP Participant",
+							Action:    transferSIPParticipant,
+							ArgsUsage: RequestDesc[livekit.TransferSIPParticipantRequest](),
```
dennwc, line 136: This API is pretty close to the ones we have on room participants in general, so maybe we should not require a JSON file here? Could be all 3 arguments instead: `--room`, `--id` (or how do we call participant identity in CLI?) and regular argument for the transfer URI.

rektdeckard: ++ on using named args. We usually use `--identity` for participants, and could map the rooms to `--from` and `--to`.

biglittlebigben: To be consistent with other commands, should we use `--room`, `--identity` `--to`? It may be confusing to have `from` be a room, and `to` a sip url

cmd/lk/sip.go
```diff
@@ -130,10 +129,18 @@ var (
...
+							Name:   "transfer",
+							Usage:  "Transfer a SIP Participant",
+							Action: transferSIPParticipant,
+							Flags: []cli.Flag{
+								roomFlag,
+								identityFlag,
+								&cli.StringFlag{
+									Name:     "to",
+									Required: true,
+									Usage:    "`SIP URL` to transfer the call to. Use 'tel:<phone number>' to transfer to a phone",
```
dennwc, line 141: Is the `tel:` prefix a part of SIP spec? Did I miss the parsing part in the `REFER` PR?

dennwc (approved):

cmd/lk/sip.go
biglittlebigben: This is supported and documented by Twilio. It's unclear if other providers support it as well. We can adjust the documentation as needed.

PR #426, opened 2024-09-21
feat(sbx): unify app commands and support template compatibility

rektdeckard:
https://linear.app/livekit/issue/DEX-766/update-cli-to-support-template-compatibility


davidzhao (approved):
lg!

PR #429, opened 2024-09-24
Revert "fix(deps): update module github.com/livekit/protocol to v1.23.0"

rektdeckard:
Reverts livekit/livekit-cli#427


davidzhao (approved):
oops, is it a breaking change?

rektdeckard:
Yup, something broke here, needs to be looked at a little closer and right this second not the time

PR #437, opened 2024-10-04
Standardize safe names on project auth

davidzhao:
the rest of tooling is set to use the standard name. doing it on the reader path tend to create inconsistencies in behavior (for example set-default did not work for me)


rektdeckard (approved):
Odd, `set-default` works for me, tested using both canonical name and url-safe names. But I suppose ultimately it is simpler to just use only url-safe names for everything

davidzhao:
I'm traced it to a reason earlier.. and I think it had to do with how the projects were set up. There were other issues with it
* adding the same project again, didn't fail, because I had an existing project where name was equal to the new would be url-safe name.
* I had a few projects that had the similar domain prefixes (not cloud), but different names. when List was using url-safe names, they were not possible to distinguish from each other.

PR #447, opened 2024-10-14
Remove deprecated create-sip-trunk

biglittlebigben:


cmd/lk/sip.go
```diff
@@ -148,17 +148,6 @@ var (
 		},
 
 		// Deprecated commands kept for compatibility
-		{
-			Hidden:   true, // deprecated: use `sip trunk create`
-			Name:     "create-sip-trunk",
-			Usage:    "Create a SIP Trunk",
-			Action:   createSIPTrunkLegacy,
```
dennwc, line 155: I think we could keep the CLI definition, but print a message pointing to `lk sip inbound create` and `lk sip outbound create`.

davidzhao:
I didn't see this PR and already merged #451

PR #451, opened 2024-10-20
Update dependencies and remove SIP.CreateSIPTrunk

davidzhao:


rektdeckard (approved):

biglittlebigben:
@dennwc had commented in a separate PR that we should throw a specific error pointing to the replacement commands. Is this something we should do indeed?

davidzhao:
I think we can go without.. I've looked at the logs and no one's used this API for awhile

PR #456, opened 2024-10-30
feat(docs+cmd): document bootstrap feature, add template list

rektdeckard:
- Bump version to `2.3.0`
- Add instructions for bootstrapping an app from template
- Prints template tags next to name when listing

Screenshot 2024-10-29 at 6 21 46 PM

- Adds `lk app list-templates` command, with table or `--json` output
image

- Shows default project with `*` instead of a whole table column dedicated to it

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,50 +97,13 @@ lk app create --template <template_name> my-app
 
 Then follow the CLI prompts to finish your setup.
 
-The `--template` flag may be omitted to see a list of all available templates, or can be chosen from a selection of our first-party templates:
+For a list of all available templates, run:
 
-<table>
-  <thead><tr><th>Template Name</th><th>Language/Framework</th><th>Description</th></tr></thead>
-  <tbody>
-    <tr>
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -86,6 +87,61 @@ lk project list
...
+```
+
+Then follow the CLI prompts to finish your setup.
+
+The `--template` flag may be omitted to see a list of all available templates, or can be chosen from a selection of our first-party templates:
+
+<table>
+  <thead><tr><th>Template Name</th><th>Language/Framework</th><th>Description</th></tr></thead>
+  <tbody>
+    <tr>
```
davidzhao, line 105: is it possible to make this markdown? would make adding entries much easier.

davidzhao: actually, do you think we need to list them? or is there a mode to print this table from livekit-cli ? `lk app list-templates`

rektdeckard: This sound much better!

davidzhao (approved):
🎉

PR #465, opened 2024-11-13
feat(cmd): allow tables to be output as json

### LINKED ISSUE #454: Feature request: support json output in list commands
Please support a `--json` flag for the `list` commands, such as `lk sip inbound list`, which currently prints a formatted table.

rektdeckard:
All commands with structured data output typically printed as a table may now be printed to stdout as JSON by using the `--json` flag. We also suppress additional output some commands would print unless used with the `--verbose` flag, in order to keep the output valid JSON.

Resolves  #454

### DIFF
--- a/cmd/lk/app.go
+++ b/cmd/lk/app.go
@@ -72,9 +72,10 @@ var (
 							Hidden:      true,
 						},
 						&cli.BoolFlag{
-							Name:   "install",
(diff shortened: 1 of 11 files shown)


davidzhao (approved):
lgtm

dennwc (approved):

PR #473, opened 2024-12-07
Add agent dispatch commands

davidzhao:
also updated to Pion v4


cmd/lk/room.go
```diff
@@ -73,8 +73,8 @@ var (
 							TakesFile: true,
 						},
 						&cli.StringFlag{
-							Name:  "room-configuration",
-							Usage: "`NAME` of the room configuration to associate with the created room",
+							Name:  "room-preset",
```
boks1971, line 76: should this a minor version bump? not sure if anybody is using this already that would fail with an upgrade, guess it won't be a silent failure, so probably okay

cmd/lk/dispatch.go
```diff
@@ -0,0 +1,214 @@
...
+	"github.com/livekit/protocol/livekit"
+	"github.com/livekit/protocol/utils"
+	lksdk "github.com/livekit/server-sdk-go/v2"
+)
+
+var (
+	DispatchCommands = []*cli.Command{
+		{
+			Name:     "dispatch",
+			Usage:    "Create, list, and delete Agent Dispatches",
```
boks1971, line 19: :nit: can all verbs have the same casing? Maybe make it `create`? feels better than making it `List` and `Delete`.

boks1971: `get` is not included in the usage here, but is there below.

boks1971: oh `get` is a specialised `list`

boks1971 (approved):

paulwe (approved):

cmd/lk/room.go
davidzhao: yeah, we'd have to do a minor bump. I don't believe anyone's using this

cmd/lk/dispatch.go
davidzhao: yes.. i'll fix the casing.. good eye!

davidzhao: I just lowercased `agent dispatches` to be consistent.. Create should be uppercased since we have been using UC for first letter of the sentence.

PR #474, opened 2024-12-09
feat(cmd): better ergonomics around `lk app` and templates

rektdeckard:
- `lk app env` command is no longer hidden
- `lk app env` now prints to stdout instead of clobbering `.env.local` unexpectedly, but can be called with the `--write` flag to keep previous behavior
- `lk app env [DIR]` is now supported, allowing you tun run outside CWD
- We no longer print an error if a template cloned using `lk app create` does not contain a `taskfile.yaml`, it is now considered optional

NOTE: we no longer recursively search the whole project for `.env.example` files -- only in the root directory. This means that we don't support "meta templates" that stitch together other templates to make a full application stack, but we were not using it anyway, and prefer to direct users to bootstrap each component separately.

### DIFF
--- a/cmd/lk/app.go
+++ b/cmd/lk/app.go
@@ -103,14 +103,20 @@ var (
 							Usage:   "Write environment variables to .env.local file",
 						},
 					},
-					Before: requireProject,
+					ArgsUsage: "[DIR] location of the project directory (default: current directory)",
+					Before:    requireProject,
 					Action: func(ctx context.Context, cmd *cli.Command) error {
-						env, err := instantiateEnv(ctx, cmd, ".", nil)
+						rootDir := cmd.Args().First()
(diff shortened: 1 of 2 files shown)


davidzhao (approved):
lg!

one question / potential feature creep: does this work when I'm outside of a bootstrapped project too? It seems that the common env vars could be useful even outside of the context of bootstrap

rektdeckard:
@davidzhao it works with any directory with a `.env.example` file 👍. It might also be nice to let this command accept an optional path argument, so you can point it to a directory besides the CWD?

rektdeckard:
Or were you saying that in the absence of a `.env.example` file, `lk app env` should just always print API Key, Secret, and URL of the default or selected project, in shell variable format, wherever you are?

PR #480, opened 2024-12-18
Add a clearer error message for invalid sandbox id

bcherry:
previously the command just said `404 Not Found` and it wasn't obvious the problem was the sandbox param, vs the template param.


rektdeckard (approved):
Thanks 🙏. Error self explanatory :)

PR #482, opened 2024-12-18
Allow custom env file paths

bcherry:
I'm working on a swift example app for the voice assistant and it needs to get the sandbox ID loaded in, but none of the API keys. The Android example app [does this with a global variable](https://github.com/livekit-examples/android-voice-assistant/blob/main/taskfile.yaml) that gets overridden via `sed`, but the approach is brittle and a little messy. 

In Swift, the best approach is to use xcconfig files, which can be imported in as build settings. It turns out that xcconfig files basically have the same format as env files, so we can just treat them as the env files for this type and take advantage of our existing env-population code. The only issues are 
 a) the files need to be in a subdirectory
 b) it's best to give them the `.xcconfig` extension (although not strictly necessary, it can work without this)
~c) we should not install the project secrets into them~
 
This PR resolves these issues by allowing you to specify custom paths to these files. I added support for these to the taskfile's vars field and also to the `app env` command for completeness.
 
 Here's a [sample taskfile](https://github.com/livekit-examples/voice-assistant-swift/blob/main/taskfile.yaml) using these new properties, which I have tested and works exactly as desired. I haven't evaluated it yet but I'm hopeful that we can adopt the same method in android, react native, and flutter.

### DIFF
--- a/cmd/lk/app.go
+++ b/cmd/lk/app.go
@@ -31,16 +31,16 @@ import (
 )
 
 var (
-	template     *bootstrap.Template
-	templateName string
-	templateURL  string
-	sandboxID    string
-	appName      string
-	appNameRegex = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_-]*$`)
-	outputFile   string
-	exampleFile  string
-	project      *config.ProjectConfig
-	AppCommands  = []*cli.Command{
+	template        *bootstrap.Template
+	templateName    string
+	templateURL     string
+	sandboxID       string
+	appName         string
+	appNameRegex    = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_-]*$`)
+	destinationFile string
+	exampleFile     string
+	project         *config.ProjectConfig
+	AppCommands     = []*cli.Command{
 		{
 			Name:     "app",
 			Category: "Core",
@@ -117,7 +117,7 @@ var (
 							Usage:       "Destination file path, when used with --write",
 							Value:       ".env.local",
 							TakesFile:   true,
-							Destination: &outputFile,
+							Destination: &destinationFile,
 						},
 						&cli.StringFlag{
 							Name:        "example",
@@ -392,7 +392,7 @@ func manageEnv(ctx context.Context, cmd *cli.Command) error {
 	}
 
 	if cmd.Bool("write") {
-		return bootstrap.WriteDotEnv(rootDir, outputFile, env)
+		return bootstrap.WriteDotEnv(rootDir, destinationFile, env)
 	} else {
 		return bootstrap.PrintDotEnv(env)
 	}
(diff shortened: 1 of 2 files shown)


rektdeckard (commented):
I wholeheartedly support the ability to specify the `--output` and `--example` input, but I should point out that instantiating the env does not *copy* LiveKit project secrets. It reads in the example and replaces only the ones present that it recognizes. So the solution to not including secrets is...not to specify them in your example.

cmd/lk/app.go
```diff
@@ -109,6 +112,24 @@ var (
 							Aliases: []string{"write"},
 							Usage:   "Write environment variables to .env.local file",
 						},
+						&cli.StringFlag{
+							Name:        "output",
+							Usage:       "Specify a custom file name for the environment variables file",
```
rektdeckard, line 117:
```suggestion
							Usage:       "Specify a custom file name for the environment variables file when used with --write",
```

bcherry:
@rektdeckard ah that makes perfect sense. it wasn't clear to me from the code, but I tested it and you are absolutely right! the only case it would fill them anyways is if no .env.example exists, but in this case I'll have an example file so it should work.

I simplified that option away and made some other tidies.

cmd/lk/app.go
```diff
@@ -291,13 +309,27 @@ func setupTemplate(ctx context.Context, cmd *cli.Command) error {
...
 	fmt.Println("Instantiating environment...")
 	addlEnv := &map[string]string{"LIVEKIT_SANDBOX_ID": sandboxID}
-	env, err := instantiateEnv(ctx, cmd, appName, addlEnv)
+	envOutputFile := ".env.local"
+	if customOutput, ok := tf.Vars.Get("env_file").Value.(string); ok {
+		envOutputFile = customOutput
+	}
+	envExampleFile := ".env.example"
+	if customExample, ok := tf.Vars.Get("env_example").Value.(string); ok {
+		envExampleFile = customExample
```
rektdeckard, line 325: Despite what it says here https://github.com/livekit-examples/index?tab=readme-ov-file#creating-your-own-template, a `taskfile.yaml` is optional and `bootstrap.ParseTaskfile` wil return `nil, nil` if none is present. I didn't check the Task docs, but just wanted to be sure there isn't a NPE waiting in this code!

cmd/lk/app.go
```diff
@@ -354,24 +386,27 @@ func manageEnv(ctx context.Context, cmd *cli.Command) error {
...
+		return bootstrap.WriteDotEnv(rootDir, outputFile, env)
 	} else {
 		return bootstrap.PrintDotEnv(env)
 	}
 }
 
-func instantiateEnv(ctx context.Context, cmd *cli.Command, rootPath string, addlEnv *map[string]string) (map[string]string, error) {
-	env := map[string]string{
+func instantiateEnv(ctx context.Context, cmd *cli.Command, rootPath string, addlEnv *map[string]string, exampleFile string) (map[string]string, error) {
+	env := map[string]string{}
```
rektdeckard, line 402: ?

rektdeckard (approved):
Other than comments, looks good!

cmd/lk/app.go
bcherry: ah missed this when clearing out the skipLiveKitKeys option I'd added. i'll clean the diff here

cmd/lk/app.go
bcherry: makes sense, fixed that and tested it too

PR #490, opened 2024-12-30
fix(cmd): relocate temp files across filesystem boundaries

rektdeckard:
https://linear.app/livekit/issue/DEX-1044/support-lk-app-create-when-run-from-non-boot-drive

- Fix failure to relocate files from `/tmp` (or equivalent) to permanent path when said path is on a different filesystem
- Move some utils to `pkg` and clean up some unused code


davidzhao (approved):
lg

PR #491, opened 2024-12-31
refactor(cmd+pkg): upgrade urfave and reorganize some code

rektdeckard:
- Upgrades `urfave/cli`
- Relocates some shared utilities to `pkg/`
- Adds some missing copyright headers
- Updates `load-test` to use fancy tables like other commands do:
Screenshot 2024-12-30 at 8 36 52 PM


cmd/lk/egress.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2023 LiveKit, Inc.
+// Copyright 2024 LiveKit, Inc.
```
davidzhao, line 1: nit: if the intent was to indicate year of modification, I would recommend using a range instead.. i.e. `2023-2024`.

changing it to the current year would lose information about when the file was originally created

rektdeckard: Fair point! I did a little git scripting and got these updated to reflect the full history of each file.

PR #494, opened 2025-01-02
Point cloud users from project add to cloud auth, remember device name

bcherry:
If you're trying to add cloud projects, the easiest way is with cloud auth. This fixes the help for the add command to tell you that.

I also noticed its annoying to retype your device name every time you auth a project, so I added it to the global config for reuse.

### DIFF
--- a/cmd/lk/cloud.go
+++ b/cmd/lk/cloud.go
@@ -269,13 +269,19 @@ func tryAuthIfNeeded(ctx context.Context, cmd *cli.Command) error {
 		return err
 	}
 
-	// name
-	var deviceName string
+	// get devicename
+	deviceName := cliConfig.DeviceName
 	if err := huh.NewInput().
 		Title("What is the name of this device?").
 		Value(&deviceName).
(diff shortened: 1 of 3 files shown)


cmd/lk/cloud.go
```diff
@@ -269,15 +269,21 @@ func tryAuthIfNeeded(ctx context.Context, cmd *cli.Command) error {
 		return err
 	}
 
-	// name
-	var deviceName string
+	// get devicename
+	deviceName := cliConfig.DeviceName
 	if err := huh.NewInput().
 		Title("What is the name of this device?").
 		Value(&deviceName).
```
rektdeckard, line 276: Nit: you can just pass in `&cliConfig.DeviceName` here and skip the intermediate string allocation

rektdeckard (approved):

cmd/lk/cloud.go
bcherry: great tip, thanks!

PR #502, opened 2025-01-17
nanpa: init at 2.3.1

nbsp:
this PR adds [nanpa](https://github.com/nbsp/nanpa) tooling to LiveKit CLI. on every new PR that contains code changes, make sure to add a file at `/.nanpa/something.kdl`, detailing the change:

```
patch type="added" "Add foo to bar"
minor type="security" "Oops"
```

multiple changes can be added per file. a future version of nanpa will add a handy CLI for adding them, for now they need to be added by hand. see [Keep A Changelog](https://keepachangelog.com/en/1.1.0/#how) for valid types.

when the [GitHub app](https://github.com/nbsp/ilo) detects changes, it will open an issue detailing them. when it is closed, it will run CI to tag versions. a release still doesn't happen automatically, in case of any bugs — you have to make a release from the tag yourself.

i have added the sample ilo-nanpa bump workflow but i don't know how the rest of the CI works here, so feel free to iterate.

@davidzhao we need to add the ilo-nanpa app to livekit-cli repository before merging, and to add `NANPA_WORKFLOW=bump.yml` to the repository variables (not secrets!)

### DIFF
--- a/.github/bump-version.sh
+++ b/.github/bump-version.sh
@@ -1,0 +1,3 @@
+#!/usr/bin/env bash
+
+sed -i "s/^\\tVersion.*\$/\tVersion = \"$VERSION\"/" version.go
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -1,0 +1,18 @@
+on:
+  workflow_dispatch:
+    inputs:
+      packages:
+        description: "packages to bump"
+        type: string
+        required: true
+
+jobs:
+  bump:
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: nbsp/ilo@v2
+        with:
+          packages: ${{ github.event.inputs.packages }}
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -16,3 +16,4 @@ jobs:
       - uses: nbsp/ilo@v2
         with:
           packages: ${{ github.event.inputs.packages }}
+          prefix: v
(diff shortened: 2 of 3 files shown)


nbsp:
i just remembered go requires `vX.Y.Z` instead of plain `X.Y.Z` as tags, marking as draft until i find a solution in ilo-nanpa

davidzhao:
> @davidzhao we need to add the ilo-nanpa app to livekit-cli repository before merging, and to add `NANPA_WORKFLOW=bump.yml` to the repository variables (not secrets!)

I'll add the app. for the WORKFLOW file.. should we just look in a default location? i.e. `nanpa.yaml`?

nbsp:
> > @davidzhao we need to add the ilo-nanpa app to livekit-cli repository before merging, and to add `NANPA_WORKFLOW=bump.yml` to the repository variables (not secrets!)
> 
> I'll add the app. for the WORKFLOW file.. should we just look in a default location? i.e. `nanpa.yaml`?

i made this an explicit setting, because the app calls `workflow_dispatch` on an action, and i don't want it to accidentally run the wrong file or to try to run a file that doesn't exist. instead it checks that `$NANPA_WORKFLOW` exists, and assumes the user was correct in setting it.

davidzhao:
> i made this an explicit setting, because the app calls `workflow_dispatch` on an action, and i don't want it to accidentally run the wrong file or to try to run a file that doesn't exist. instead it checks that `$NANPA_WORKFLOW` exists, and assumes the user was correct in setting it. the workflow added in this PR is `bump.yml`.

picking a sane default would help to reduce the # of steps that a user would have to perform. While it's a fair concern that a user may have a collision with another file intended for a different purpose, it seems that we can prevent it by validating that it is indeed an nanpa file. just my 2c. I've added this env var in any case

davidzhao (approved):

PR #522, opened 2025-02-13
Don't copy LICENSE or NOTICE files with bootstrap

bcherry:
When a customer uses the bootstrap function to clone an example app, we intend to give them complete rights to the resulting code to basically do whatever they want.

By including the LICENSE file, we are actually causing them to apply it (whether intentional or not) if they publish it. So let's not copy it.

There's maybe a deeper issue worth considering which has to do with copyright. The Apache 2 license requires a copyright NOTICE in the repo as well as a copyright notice at the top of every source file. But it doesn't really make sense for us to copyright bootstrap templates in this way... It would be ideal if we could strip the copyright notices from the files as well or simply choose an even less restrictive license for our sample apps such as MIT.  And that could also be a requirement enforced by the tool - perhaps it should refuse to work on a repository not covered by the MIT license or similar.

In any case, this is the simple step and we can think about the rest later.

### DIFF
--- a/.nanpa/no-license.kdl
+++ b/.nanpa/no-license.kdl
@@ -1,0 +1,1 @@
+patch type="added" "Do not copy LICENSE or NOTICE files during bootstrap"
--- a/pkg/bootstrap/bootstrap.go
+++ b/pkg/bootstrap/bootstrap.go
@@ -62,6 +62,8 @@ var templateIgnoreFiles = []string{
 	"renovate.json",
 	"taskfile.yaml",
 	"TEMPLATE.md",
+	"LICENSE",
+	"LICENSE.md",
 }
 
 type Template struct {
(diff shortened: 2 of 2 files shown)


chrisackermann (approved):
🙏

davidzhao (approved):
makes sense.

we should change the license of our templates to MIT or something that doesn't require attribution

bcherry:
@davidzhao yeah you think there's anything else to consider before we just do that? happy to start changing them

davidzhao:
we are good to change it. the spirit of those licenses is to make it as permissible as possible.

bcherry:
cool I switched all of the examples that are explicitly templates or starter apps over to MIT but left others that are "whole apps" alone (meet etc)

PR #530, opened 2025-03-17
add cloud agents

real-danm:
related:
https://github.com/livekit/protocol/pull/1010
https://github.com/livekit/server-sdk-go/pull/625


cmd/lk/agent.go
```diff
@@ -0,0 +1,1040 @@
...
+			return fmt.Errorf(fmt.Sprintf("%s required to create agent", cmd.String("toml")))
+		}
+
+		f, err := os.Create(filepath.Join(workingDir, cmd.String("toml")))
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+
+		subdomainPattern := regexp.MustCompile(`^(?:https?|wss?)://([^.]+)\.`)
```
paulwe: isn't this already defined?

paulwe: isn't this already defined

pkg/agentfs/docker.go
```diff
@@ -0,0 +1,265 @@
...
+
+	"github.com/livekit/protocol/logger"
+)
+
+const (
+	pythonDockerfileURL   = "https://raw.githubusercontent.com/livekit-examples/agent-deployment/refs/heads/main/python-agent-example-app/Dockerfile"
+	pythonDockerIgnoreURL = "https://raw.githubusercontent.com/livekit-examples/agent-deployment/refs/heads/main/python-agent-example-app/.dockerignore"
+	pythonEntrypoint      = "main.py"
+	nodeDockerfileURL     = "https://raw.githubusercontent.com/livekit-examples/agent-deployment/refs/heads/main/node-agent-example-docker/Dockerfile"
+	nodeDockerIgnoreURL   = "https://raw.githubusercontent.com/livekit-examples/agent-deployment/refs/heads/main/node-agent-example-docker/.dockerignore"
```
paulwe, line 23: are we confident these aren't going to change? should we load them from a config endpoint in the api?

paulwe (approved):

pkg/agentfs/docker.go
real-danm: Yeah, that's reasonable, it will probably be helpful to have a generic config endpoint for a few different things anyways, will add this in.

pkg/agentfs/build.go
```diff
@@ -87,7 +90,7 @@ func Build(ctx context.Context, id string, name string, action string, projectCo
 		for scanner.Scan() {
 			line := scanner.Text()
 			if strings.HasPrefix(line, "BUILD ERROR:") {
-				return fmt.Errorf(strings.TrimPrefix(line, "BUILD ERROR: "))
+				return fmt.Errorf("%s", strings.TrimPrefix(line, "BUILD ERROR: "))
```
paulwe, line 93: `errors.New`?

pkg/agentfs/build.go
```diff
@@ -0,0 +1,114 @@
...
+		_, err = display.UpdateFrom(context.Background(), ch)
+		return err
+	})
+
+	eg.Go(func() error {
+		defer close(ch)
+		scanner := bufio.NewScanner(resp.Body)
+		for scanner.Scan() {
+			line := scanner.Text()
+			if strings.HasPrefix(line, "BUILD ERROR:") {
```
boks1971, line 92: is there a more structured way to get build errors? Or is this  always reliable?

pkg/agentfs/build.go
```diff
@@ -0,0 +1,114 @@
+package agentfs
```
boks1971, line 1: what does `fs` stand for?

boks1971 (approved):
could not read all of it, but lgtm as much as I read!

pkg/agentfs/build.go
real-danm: agent-filesystem.. was trying to be a bit generic since there's a lot of functionality in this package.

pkg/agentfs/build.go
real-danm: The server side is actively setting this, it's a streaming endpoint so setting http codes doesn't work.

PR #543, opened 2025-04-06
fix symlink file handling and add tests for tarball create

real-danm:
this fixes the tarball create issues with symlinks
```
failed to walk directory: failed to copy file content for res/agent_male.png: current number exceeds max
```


rektdeckard (approved):
Note to self: test on Windows

PR #544, opened 2025-04-07
Add a flag for using dev credentials

dennwc:
This change adds `--dev` flag that uses the same credentials as LiveKit server's `--dev` flag.


rektdeckard (approved):
I'm a little wary this doesn't override nicely when used with `--project` or other plays that affect which project config is loaded and what environment is loaded. It feels like it might need to be mutually exclusive, or at least be checked in each of the early returns in loadProjectConfig?

PR #549, opened 2025-04-13
added agent load testing feature

s-hamdananwar:


pkg/loadtester/agentloadtester.go
```diff
@@ -0,0 +1,335 @@
...
+
+			_, err := loadTestRoom.publishEchoTrack()
+			if err != nil {
+				log.Printf("Failed to publish echo track to room %s: %v", roomName, err)
+				loadTestRoom.stop()
+				return err
+			}
+			loadTestRoom.stats.echoTrackPublished = true
+
+			err = loadTestRoom.dispatchAgent()
```
rektdeckard, line 103: Curious why agent needs to be manually dispatched to the room? Won't this happen automatically?

rektdeckard (commented):
Overall looks sound to me, but in testing against a locally-running agent, I had join rates of only 0-40%, and not once did all rooms get joined. I suspect there may be a race condition here, might be due to creating the room and then manually dispatching to a room with the same name in such short time? I know the the router is supposed to auto-allocate a room when a dispatch is made to a room that does not exist. Is it possible agents are sent to a room with the same name, but actually not the same room?

Screenshot 2025-04-14 at 3 32 33 PM
Screenshot 2025-04-14 at 3 32 41 PM
Screenshot 2025-04-14 at 3 32 49 PM

s-hamdananwar:
@rektdeckard were you running the agent using `dev`? If so, do you notice any difference in joining rates when using `start`? I had similar results when running with `dev` and found it to be solved when using `start`. I think this is an Agents issue, specifically the job accepting implementation difference between the two modes. If that is indeed the issue, maybe I also should instruct to use `start` in the readme NOTE ?

s-hamdananwar:
Also, thanks for the screenshots- added a quick check to make sure to not report false `Agent Track Subscribed` values.

rektdeckard (approved):
Yep, turned out `start` worked perfectly. Nicely done! Let's add a callout to that in the readme for sure.

One other small point, for the sake of organization is there a better place for this command than at the root? Thinking it might be forward-thinking to do:

```
lk
└─ perf
   └─ load-test
    └─ agent-load-test
```

We can leave the `load-test` command aliased at the root so we don't break any existing workflows too. Feel free to implement or not, I'm happy to come back to it before we release again.

rektdeckard:
Explained in slack, but my miscommunication, I was talking about the CLI command structure, not the Go package structure.

pkg/loadtester/agentloadtester.go
```diff
@@ -0,0 +1,337 @@
...
+		log.Printf("Failed to join room %s (attempt %d): %v", roomName, i+1, err)
+		time.Sleep(1 * time.Second)
+	}
+	if err != nil {
+		return err
+	}
+
+	meetParticipantToken, _ := newAccessToken(r.params.APIKey, r.params.APISecret, roomName, "meet-participant")
+	r.stats.meetLink = fmt.Sprintf("https://meet.livekit.io/custom?liveKitUrl=%s&token=%s", r.params.URL, meetParticipantToken)
+	log.Printf("\nInspect the room %s in LiveKit Meet using this url: %s\n", roomName, r.stats.meetLink)
```
davidzhao, line 181: this is neat! though would this print potentially 100 times if you are starting 100 agents? that'd be kind of verbose.. perhaps hide behind verbose flag?

pkg/loadtester/agentloadtester.go
```diff
@@ -0,0 +1,337 @@
...
+		Name: "echo-track",
+	})
+	if err != nil {
+		return "", err
+	}
+	return p.SID(), nil
+}
+
+func (r *LoadTestRoom) dispatchAgent() error {
+	dispatchClient := lksdk.NewAgentDispatchServiceClient(r.params.URL, r.params.APIKey, r.params.APISecret)
```
davidzhao, line 207: if the user didn't pass in an agent name.. that would just be the default agent. we need to skip dispatch here, otherwise we'd end up with two agents

pkg/loadtester/agentloadtester.go
```diff
@@ -0,0 +1,337 @@
...
+		return err
+	}
+	r.stats.agentDispatchedAt = time.Now()
+	// log.Printf("Successfully dispatched agent %s to room %s", r.params.AgentName, r.room.Name())
+	return nil
+}
+
+func (r *LoadTestRoom) onTrackSubscribed(track *webrtc.TrackRemote, pub *lksdk.RemoteTrackPublication, rp *lksdk.RemoteParticipant) {
+	if r.echoTrack != nil && r.firstParticipant == nil && track.Kind() == webrtc.RTPCodecTypeAudio {
+		log.Printf("Subscribing to echo track in room %s, delay: %s", r.room.Name(), r.params.EchoSpeechDelay.String())
```
davidzhao, line 224: would be better to print delay/config value once.. versus on each track

pkg/loadtester/agentloadtester.go
```diff
@@ -0,0 +1,337 @@
...
+
+func (t *AgentLoadTester) printStats() {
+	t.lock.Lock()
+	defer t.lock.Unlock()
+
+	checkStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("2")) // Green
+	crossStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("1")) // Red
+
+	table := util.CreateTable().
+		Headers("#", "Room", "Agent Dispatched At", "Agent Joined", "Agent Joined At", "Agent Track Subscribed", "Echo Track Published")
```
davidzhao, line 299: instead of agent joined at.. would be great to display agent join delay. how long did it take the agent to join?

PR #553, opened 2025-04-17
Adding support for room join with attributes

wutiphong-sir:
for room join this will help setting attributes for participant with 2 new args
--attribute <key=value>
--attribute-file <JSON File>

### DIFF
--- a/.nanpa/room-join-attr.kdl
+++ b/.nanpa/room-join-attr.kdl
@@ -1,0 +1,1 @@
+minor type="added" "Added room joining with setting attributes and attribute file"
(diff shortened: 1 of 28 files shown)


rektdeckard (changes requested):
Nice addition! Can you remove the changes in `pkg/provider/resources`? These are test files stored with Git LFS, you don't need them to make your contribution but please don't delete them :)

wutiphong-sir:
@rektdeckard Thank you for reviewing. The deleted test files has been restored already. (I somehow unintentionally removed it, apologize for that)

PR #554, opened 2025-04-18
fix(agents): correct logic for agents secrets flags

rektdeckard:
https://linear.app/livekit/issue/HA-95/secret-file-required-even-when-providing-secrets

Fix logic around agent secrets:
- None required for `lk agent deploy`
- Use lazy exaluation if secrets set via flags in `lk agent update-secrets`

### DIFF
--- a/cmd/lk/agent.go
+++ b/cmd/lk/agent.go
@@ -68,9 +68,10 @@ var (
 	}
 
 	secretsFileFlag = &cli.StringFlag{
-		Name:     "secrets-file",
-		Usage:    "`FILE` containing secret KEY=VALUE pairs, one per line. These will be injected as environment variables into the agent.",
-		Required: false,
+		Name:      "secrets-file",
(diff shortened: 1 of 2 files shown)


cmd/lk/agent.go
```diff
@@ -522,19 +524,21 @@ func deployAgent(ctx context.Context, cmd *cli.Command) error {
...
 		MaxReplicas: int32(agentConfig.MaxReplicas),
 	}
 
+	secrets, err := requireSecrets(ctx, cmd, false, true)
+	if err != nil {
+		return err
+	}
+	if len(secrets) > 0 {
+		req.Secrets = secrets
+	}
```
rektdeckard, line 540: @real-danm just check my math here, it's totally fine to deploy with no secrets right, because we assume there were valid secrets in the initial creation?

real-danm: yep correct

real-danm (approved):

PR #566, opened 2025-05-05
add --auto-subscribe cli option to enable auto-subscribe to published tracks in a room

chenosaurus:
When using LK CLI to directly publish tracks into a room, it defaults to also auto-subscribing to other participant's tracks.  This is undesirable as most use cases of the CLI does not need to receive any tracks and will waste bandwidth.  This adds a `--auto-subscribe` parameter to the CLI when using the join room functionality to explicitly enable auto-subscribing to published tracks.


cmd/lk/room.go
```diff
@@ -177,6 +177,11 @@ var (
 							Usage:     "read attributes from a `JSON` file",
 							TakesFile: true,
 						},
+						&cli.BoolFlag{
+							Name:  "auto-subscribe",
+							Usage: "Automatically subscribe to published tracks. Default is false.",
+							Value: false,
```
bcherry, line 183: i'm hesitant to change the default behavior in this PR. maybe best as true

bcherry (approved):

cmd/lk/room.go
```diff
@@ -177,6 +177,11 @@ var (
 							Usage:     "read attributes from a `JSON` file",
 							TakesFile: true,
 						},
+						&cli.BoolFlag{
+							Name:  "auto-subscribe",
+							Usage: "Automatically subscribe to published tracks. Default is true.",
+							Value: true,
```
davidzhao, line 183: it should probably be false by default. the CLI client cannot really consume any tracks

davidzhao (approved):

cmd/lk/room.go
rektdeckard: Seconded. Also, when `Value` is set on a flag, the CLI will display `(defaault: $value)` so you can remove it from the usage string.

cmd/lk/utils.go
```diff
@@ -142,6 +142,14 @@ func extractFlagOrArg(c *cli.Command, flag string) (string, error) {
 	return value, nil
 }
 
+func extractFlagAsBool(c *cli.Command, flag string) (bool, error) {
+	if c.IsSet(flag) {
+		return c.Bool(flag), nil
+	}
+
+	return false, nil
+}
```
rektdeckard, line 151: I don't see anything that could possibly error here, so it doesn't need to return one

cmd/lk/room.go
```diff
@@ -784,6 +789,11 @@ func joinRoom(ctx context.Context, cmd *cli.Command) error {
 		return err
 	}
 
+	autoSubscribe, err := extractFlagAsBool(cmd, "auto-subscribe")
+	if err != nil {
+		return err
+	}
```
rektdeckard, line 795: I think we can just read `cmd.Bool("auto-subscribe")` directly, since we are setting a default value for it.

rektdeckard (commented):
Just some minor changes based on how we normally do CLI flags!

version.go
```diff
@@ -15,5 +15,5 @@
 package livekitcli
 
 const (
-	Version = "2.4.5"
+	Version = "2.4.6"
```
rektdeckard, line 18: I think you need a rebase, we're at `2.4.6` already

rektdeckard (approved):

PR #569, opened 2025-05-09
feat(lk): respect config files globally

rektdeckard:
- Now recognizes the `--config` flag globally, instead of the `--toml` flag previously used in `lk agent` commands, to point to a configuration toml file (default `livekit.toml`)
- If not otherwise specified, uses this project for commands, meaning by default, if the config file exists in the working dir and the project specified by its `[project.subdomain]` is in the "keychain", these credentials will be used
- Support new `livekit.toml` format, with backwards compatibility with the old format:

```toml
[project]
  subdomain = "my-subdomain123"

[agent]
  name = "my-agent"
  cpu = "1"
  replicas = 1
  max_replicas = 10
```

The new format gives us some comfy wiggle room if we want to use this file for other features in the future.


davidzhao (approved):
lgtm

PR #576, opened 2025-05-15
Deprecate hosted agent name from config

yoonsio:
All hosted agents are identified via unique ID instead of user-defined name.

* Use agent ID for all hosted agent API.
* Automatically update livekit.toml file with agent ID.

### DIFF
--- a/cmd/lk/agent.go
+++ b/cmd/lk/agent.go
@@ -452,7 +452,7 @@ func createAgentConfig(ctx context.Context, cmd *cli.Command) error {
 	regionAgent := agent.AgentDeployments[0]
 	lkConfig := config.NewLiveKitTOML(matches[1])
 	lkConfig.Agent = &config.LiveKitTOMLAgentConfig{
-		Name:        agent.AgentName,
+		ID:          agent.AgentId,
 		CPU:         config.CPUString(regionAgent.CpuReq),
 		Replicas:    int(regionAgent.Replicas),
 		MaxReplicas: int(regionAgent.MaxReplicas),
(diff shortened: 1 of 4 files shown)


cmd/lk/agent.go
```diff
@@ -43,17 +41,17 @@ const (
...
 	}
 
-	nameSliceFlag = &cli.StringSliceFlag{
-		Name:     "name",
-		Usage:    "`NAMES` of agent(s)",
+	idSliceFlag = &cli.StringSliceFlag{
+		Name:     "id",
+		Usage:    "`IDs` of agent(s)",
 		Required: false,
 	}
```
rektdeckard, line 56: Do we want to preserve but deprecate the `name` flag and mark it with `Hidden: true`? Or are we just ripping the bandaid on this one since it's in beta?

rektdeckard (approved):
Looks good, just iffy on the breaking change but I think we can live with it 👍

cmd/lk/agent.go
yoonsio: What does `Hidden: true` do? It might get confusing since when user sets a name flag and notices it does nothing.

yoonsio: What does `Hidden: true` do? It might get confusing since when user sets a name flag and notices it does nothing.

rektdeckard: Oh, I meant keeping both sets of logic around for a while and actually handling the `name` parameter. But we don't need to do that.

`Hidden: true` removes the flag from the `--help` commands and other places so it isn't shown to users, but still works.

yoonsio: Ah, gotcha. The Agents API will reject any requests that depend on agent names and return an error message asking users to upgrade livekit-cli.

PR #583, opened 2025-05-23
Allow setting and updating destination-country on outbound trunk

biglittlebigben:


cmd/lk/sip.go
```diff
@@ -643,6 +654,9 @@ func updateSIPOutboundTrunk(ctx context.Context, cmd *cli.Command) error {
 		tr := livekit.SIPTransport(trv)
 		req.Transport = &tr
 	}
+	if val := cmd.String("destination-country"); val != "" {
+		req.DestinationCountry = &val
```
boks1971, line 658: Is this like the 2-digit country code? Is there some validation? Maybe, worth adding in help the format needed for this.

boks1971 (approved):

cmd/lk/sip.go
biglittlebigben: Updated the documentation

PR #584, opened 2025-05-24
refactor: embed agent docker files

rektdeckard:
https://linear.app/livekit/issue/HA-76/move-generated-dockerfiles-to-templates-in-livekit-cli

The `Dockerfile` and `.dockerignore` files required to deploy a cloud agent are now embedded in the binary, rather than fetched at runtime from [a git repo](https://github.com/livekit-examples/agent-deployment/tree/main/python-agent-example-app) based on the path defined in a `cloud-agents` [API](https://github.com/livekit/cloud-agents/blob/main/pkg/config/config.go#L101-L119).

Current CLI versions depend on that API, se we should probably leave it and the files they point to in place for a while. Unless we think it's ok to break since this is still only accessible to beta users?


real-danm:
>Current CLI versions depend on that API, se we should probably leave it and the files they point to in place for a while. Unless we think it's ok to break since this is still only accessible to beta users?

i think it's ok to remove. livekit-cli now warns users to upgrade if they're behind and we've also been announcing to the channel when there is a breaking change.

PR #598, opened 2025-06-13
additional test for compressed file uploads

real-danm:
ensures python submodule files and folders are handled correctly in compression and upload for agents


rektdeckard (approved):
So RE: that community comment, the issue was just to do with build secrets?

real-danm:
> So RE: that community comment, the issue was just to do with build secrets?

looks like it's actually an issue with windows and how it's handling the file compression

PR #600, opened 2025-06-16
fix(bootstrap): if not already, prompt to auth when bootstrapping sbx

rektdeckard:
https://linear.app/livekit/issue/DEX-1617/prompt-to-authenticate-cli-when-bootstrapping-a-sandbox

If you try to bootstrap a project without first authenticating the CLI, we should push automatically to CLI auth flow instead of failing.


Ocupe (approved):
👍

PR #603, opened 2025-06-26
chore: update the autocomplete scripts to urfave/cli v3 format

dwayn:
Our bash completion scripts are still using the urfave/cli v2 format that uses a flag `--generate-bash-completion`, but that is not implemented in v3, but instead uses the hidden subcommand `completion bash|fish|zsh` to achieve the same effect. 

This updates all the autocomplete scripts to the new implementation.


rektdeckard (approved):
Looks great. Nice catch!

rektdeckard:
@dwayn can we merge this?

PR #604, opened 2025-06-26
Enable LK CLI to publish multiple H264 streams as a simulcast track

chenosaurus:
### New features:
- allow up to 3 `--publish` flags to be passed to the CLI with the syntax `h264://<host:port>/<width>x<height>` or `h264://<socket_path>/<width>x<height>` to publish a single simulcast video track.

### Notes:
- Only H264 video streams are supported
- The height/width passed in must match the video resolution

### DIFF
--- a/cmd/lk/join.go
+++ b/cmd/lk/join.go
@@ -22,7 +22,10 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"sort"
+	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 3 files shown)


cmd/lk/join.go
```diff
@@ -368,3 +371,176 @@ func publishReader(room *lksdk.Room,
...
+	})
+	if err != nil {
+		// Clean up tracks on publish failure
+		for _, track := range tracks {
+			track.Close()
+		}
+		return fmt.Errorf("failed to publish simulcast track: %w", err)
+	}
+
+	fmt.Printf("Successfully published H.264 simulcast track with qualities: %v\n", trackNames)
```
rektdeckard, line 544: It's not necessary, but a table of the published tracks and their properties might be nice here.

cmd/lk/room.go
```diff
@@ -794,6 +800,38 @@ func _deprecatedUpdateRoomMetadata(ctx context.Context, cmd *cli.Command) error
...
+	publishUrls := cmd.StringSlice("publish")
+
+	// Determine simulcast mode by checking if any URL has simulcast format
+	simulcastMode := false
+	for _, url := range publishUrls {
+		if simulcastURLRegex.MatchString(url) {
+			simulcastMode = true
+			break
+		}
+	}
```
rektdeckard, line 812: This looks like it doesn't support mixed content of several simulcast layers PLUS other unicast content. Is that worth calling out, or expected?

rektdeckard (approved):
Looks great! Small comments, and I would also add an example to the README right below https://github.com/livekit/livekit-cli#publish-media-files, since this is pretty cool and has been asked for by several people in the past.

rektdeckard (approved):

cmd/lk/room.go
rektdeckard: Also FWIW if you want you can put flag validation logic in the flag's Action: [urfave/cli@main/docs/v3/examples/flags/advanced.md#flag-actions](https://github.com/urfave/cli/blob/main/docs/v3/examples/flags/advanced.md?rgh-link-date=2025-06-27T18%3A50%3A50Z#flag-actions). Might not be so useful here, but nice when you want to reuse flags.

chenosaurus: Yes, if using simulcast, it must only be simulcast layers without any other content.  I'll update the readme to clearly explain.

cmd/lk/join.go
chenosaurus: @rektdeckard do we have any examples in other tools that prints a table?

rektdeckard: Yeah, we use it in tons of places in the CLI, for example when you run `lk project list`: https://github.com/livekit/livekit-cli/blob/b01c31b12bc6163528bd67298d97f15b5101c849/cmd/lk/project.go#L254-L275

PR #607, opened 2025-07-01
feat: support `--open` flag and template strings

rektdeckard:
Ticket: https://linear.app/livekit/issue/DEX-1670/add-option-for-random-room-name-in-cli-token-generation

- Now supports `lk token create --open meet`, which will open Meet and automatically join the room with the generated token
- Supports template strings in `--room`, `--identity` parameters in some cases. This lets you generate pseudo-random names easily:

```shell
lk token create --join \
  --identity "tobias-%h" \
  --room "room-%t"
```


cmd/lk/token.go
```diff
@@ -84,20 +87,15 @@ var (
 							Name:  "allow-source",
 							Usage: "Restrict publishing to only `SOURCE` types (e.g. --allow-source camera,microphone), defaults to all",
 						},
-						&cli.StringFlag{
+						&TemplateStringFlag{
 							Name:    "identity",
 							Aliases: []string{"i"},
-							Usage:   "Unique `ID` of the participant, used with --join",
+							Usage:   "Unique `ID` of the participant, used with --join (supports templates)",
 						},
```
s-hamdananwar: qq: are we not reusing `optional(identityFlag)` here because we need to add more details to the usage description?

cmd/lk/token.go
```diff
@@ -324,12 +318,12 @@ func createToken(ctx context.Context, c *cli.Command) error {
...
 	}
 
-	pc, err := loadProjectDetails(c, ignoreURL)
+	_, err := requireProjectWithOpts(ctx, c, ignoreURL)
 	if err != nil {
 		return err
 	}
 
-	at := accessToken(pc.APIKey, pc.APISecret, grant, p)
+	at := accessToken(project.APIKey, project.APISecret, grant, participant)
```
s-hamdananwar, line 322: I really like this

s-hamdananwar (approved):
lgtm! super cool

cmd/lk/token.go
rektdeckard: Flags are optional by default if not specified. We use the `optional(roomFlag)` construction above because the `roomFlag` is marked required and we don't want that here, so need to make an optional copy.

s-hamdananwar: makes sense, thanks

rektdeckard: Sorry I must have had an aneurism when reading the initial comment. I thought you were talking about `roomFlag`, not `identityFlag`. You're right, we can reuse it.

PR #608, opened 2025-07-02
allow metadata to be passed for joining participant

terranvigil:
This change allows `metadata` to be passed as part of `lk room join`.

See the related Slack thread here:
https://livekit-users.slack.com/archives/C025KM0S1CK/p1750953266259539


terran-vigil-simplisafe:
I've signed the CLA. Anything else needed?

PR #612, opened 2025-07-09
support multi-region for cloud-agents

real-danm:
adds a `--region` string slice option to agent create and agent update commands. Default is empty which will deploy cloud agent to the nearest region. Supports multiple regions by using a comma separated list.

### DIFF
--- a/cmd/lk/agent.go
+++ b/cmd/lk/agent.go
@@ -75,6 +75,12 @@ var (
 		Required: false,
 	}
 
+	regionFlag = &cli.StringSliceFlag{
+		Name:     "regions",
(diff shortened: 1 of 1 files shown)


cmd/lk/agent.go
```diff
@@ -75,6 +75,12 @@ var (
 		Required: false,
 	}
 
+	regionFlag = &cli.StringSliceFlag{
+		Name:     "regions",
+		Usage:    "Region(s) to deploy the agent to. If unset, will deploy to the nearest region.",
+		Required: false,
```
rektdeckard, line 81: If we're still not sure if we want to expose this to customers yet, let's set this to `Hidden` for now

PR #618, opened 2025-07-22
ensure agent sdks on cloud-agent builds meet minimum sdk support

real-danm:
- we are setting the minimum support agent sdk versions for cloud agents server side in the client settings
- using `github.com/google/osv-scanner` to do the dependency lock file scanning, supports several different node lockfiles and python lockfiles. this only increases the size of the built lk-cli binary by ~1M
- adds tests to ensure check is working as expected
- error output shows `package livekit-agents version 1.1.6 is too old, please upgrade to 1.2.0`


kuba-- (approved):

pkg/agentfs/sdk_version_check.go
```diff
@@ -0,0 +1,68 @@
...
+	minimum, err := semver.NewVersion(minVersion)
+	if err != nil {
+		return fmt.Errorf("invalid minimum version format: %s", minVersion)
+	}
+
+	if current.LessThan(minimum) {
+		return fmt.Errorf("package %s version %s is too old, please upgrade to %s", packageName, currentVersion, minVersion)
+	}
+	return nil
+}
```
rektdeckard: I can't tell from the code, but this should probably support bare package specifiers, since they will just be installed with latest versions. This *should* be valid:

```
livekit-agents
```

rektdeckard (approved):
Looks good. In general I'm a bit wary of stuff like this, since the package files and version specifiers can get quite nuanced. For example, both `package.json` and `requirements.txt` can source other dependency files. But not sure our builds support that anyway 🤷

pkg/agentfs/sdk_version_check.go
real-danm: good catch, updated to support this

real-danm:
> Looks good. In general I'm a bit wary of stuff like this, since the package files and version specifiers can get quite nuanced. For example, both `package.json` and `requirements.txt` can source other dependency files. But not sure our builds support that anyway 🤷

I added tests for these scenarios: python nested files and workspaces in node. The osv-scanner supports both cases out of the box; this is partly why I used it rather than rolling our own, it seems pretty robust. If there are more complex cases, we may have to handle those as they get reported.

PR #623, opened 2025-07-25
feat(agents): support uv and other python project detection/setup

rektdeckard:
- Adds support for `uv` projects out of the box, including template `Dockerfile` and `.dockerignore`
- Supports entrypoint detection in nested directories. Before we only looked for `.py`/`.js` files in the root, now we recurse making it possible to detect the common case of an entrypoint in `src/`, e.g.
- Adds a new convention for dockerfiles: `ARG PROGRAM_MAIN="<entrypoint.py>"` which makes it more convenient to point to the agent entrypoint in multiple places (and for end user to override)

Oops, I guess this does most of what is in this ticket (minus documentation): https://linear.app/livekit/issue/HA-235/lk-cli-cleanup-and-dockerfile-changes


pkg/agentfs/examples/python.uv.dockerignore
```diff
@@ -0,0 +1,19 @@
...
+.vscode
+*.pyc
+*.pyo
+*.pyd
+.git
+.gitignore
+README.md
+LICENSE
+.github 
+tests/
```
real-danm: nit: new line

real-danm (approved):
lgtm!

pkg/agentfs/utils.go
```diff
@@ -23,18 +24,67 @@ import (
...
-	return false
+}
+
+func isPythonPip(dir string) bool {
+	_, err := os.Stat(filepath.Join(dir, "requirements.txt"))
+	return err == nil
+}
+
+func isPythonUV(dir string) bool {
+	_, err := os.Stat(filepath.Join(dir, "pyproject.toml"))
```
bcherry: i'm not totally sure this is the right way to detect UV. it's also possible to use pip with pyproject.toml and I think it's actually more recommended these days. a more robust way to detect uv is probably the presence of `uv.lock` instead

rektdeckard: that makes sense, although i guess we will have trouble detecting our own template since we don't commit the lockfile!

bcherry: yeah but once the dev does `uv sync` they'll get a lockfile (whether committed or not), and i'm pretty sure the docker build won't work until they do anyways because its a locked install

rektdeckard: totally right. i'll update

rektdeckard: this begs the question, how do we detect pip projects now? 😆 
`if (hasFile("pyproject.toml") || hasFile("requirements.txt")) && !isPythonUV()`?

bcherry: yeah i think so. maybe really just `!isPythonUV()` is the only thing

rektdeckard: we do have to eliminate Node projects too though, but I guess it's just a matter of ordering.

pkg/agentfs/utils.go
```diff
@@ -15,26 +15,99 @@
...
 
-func isPython(dir string) bool {
-	if _, err := os.Stat(filepath.Join(dir, "requirements.txt")); err == nil {
-		return true
+type ProjectType string
+
+const (
+	ProjectTypePythonPip ProjectType = "python.pip"
+	ProjectTypePythonUV  ProjectType = "python.uv"
+	ProjectTypeNode      ProjectType = "node"
```
bcherry, line 34: doesn't have to be in this PR but tbh we should probably also detect pnpm (pnpm-lock.yaml), npm (package-lock.json) and yarn (yarn.lock) for node

rektdeckard: @dwayn is absorbing this work into his PR over here: https://github.com/livekit/livekit-cli/pull/624

cmd/lk/agent.go
```diff
@@ -369,15 +376,24 @@ func createAgent(ctx context.Context, cmd *cli.Command) error {
...
+	projectType, err := agentfs.DetectProjectType(workingDir)
+	if err != nil {
+		return fmt.Errorf("unable to determine project type: %w, please use a supported project type, or create your own Dockerfile in the current directory", err)
+	}
+
+	if err := requireDockerfile(ctx, cmd, workingDir, projectType, settingsMap); err != nil {
 		return err
 	}
 
 	// TODO (steveyoon): disable check SDK version until we add support for uv and lockfile generation
```
real-danm: can remove these two TODO lines as well

pkg/agentfs/docker.go
```diff
@@ -93,67 +88,68 @@ func CreateDockerfile(dir string, settingsMap map[string]string) error {
 	return nil
 }
 
-func validateEntrypoint(dir string, dockerfileContent []byte, projectType string, settingsMap map[string]string) ([]byte, error) {
-	fileList := make(map[string]bool)
-	entries, err := os.ReadDir(dir)
-	if err != nil {
-		return nil, err
-	}
+func validateEntrypoint(dir string, dockerfileContent []byte, dockerignoreContent []byte, projectType ProjectType, settingsMap map[string]string) ([]byte, error) {
```
real-danm, line 91: maybe need some tests for a few functions in this file?

real-danm (approved):
optional for some tests for the pkg/agentfs/docker.go but otherwise lgtm!

pkg/agentfs/docker.go
rektdeckard: This function takes user input and renders a TUI picker, I'll split it up and test some smaller bits in a follow-up PR?

PR #625, opened 2025-07-26
feat(apps): officially support remote taskfiles

rektdeckard:
This patch adds official support for remote taskfiles in templates, so we can define common bootstrap tasks in the [index](https://github.com/livekit-examples/index/blob/main/taskfile.yaml) or elsewhere.


bcherry (approved):
this reminds me that IMO we should delete the taskfile from the destination folder during copy

rektdeckard:
@bcherry we do that already: https://github.com/livekit/livekit-cli/blob/main/pkg/bootstrap/bootstrap.go#L268-L276

bcherry:
@rektdeckard oh - weird because i swear this happened to me just last week

PR #636, opened 2025-08-07
Handle agent delete error

yoonsio:


cmd/lk/agent.go
```diff
@@ -785,27 +785,25 @@ func deleteAgent(ctx context.Context, cmd *cli.Command) error {
...
-	if innerErr != nil {
-		if twerr, ok := err.(twirp.Error); ok {
+	if agentErr != nil {
+		if twerr, ok := agentErr.(twirp.Error); ok {
 			if twerr.Code() == twirp.PermissionDenied {
 				return fmt.Errorf("agent hosting is disabled for this project -- join the beta program here [%s]", cloudAgentsBetaSignupURL)
 			}
 		}
-		return err
+		return agentErr
```
rektdeckard, line 806: In this case should we maybe not return the error at all? Below we check `res.Success`, and maybe its okay to swallow an error if it was not catastrophic? Unsure if that can even happen, but just thinking out loud. We don't want to lead user to believe the agent wasn't deleted if it actually succeeded.

yoonsio: Looking at the server code, the error always indicates that the agent was not deleted. Not sure why the error was swallowed earlier when we check `res.Success`

PR #642, opened 2025-08-11
chore(agents): fallback to agent selector when generating config

rektdeckard:
- Allow user to select existing Agent ID if not provided in `lk agent config`
- Respects `workingDir` when creating config file
- Bump version


yoonsio (approved):
Thanks!

PR #643, opened 2025-08-11
fix(agents): correctly expand `PROGRAM_MAIN` in Dockerfiles

rektdeckard:
Fix expansion of the `PROGRAM_MAIN` variable in python Dockerfile templates by using native Golang string templates. This vastly simplifies the previous entrypoint detection we were doing previously.

Have tested this with both a `pip` and `uv` project locally, as well as with cloud deployments and it works 👍

### DIFF
--- a/pkg/agentfs/examples/python.pip.Dockerfile
+++ b/pkg/agentfs/examples/python.pip.Dockerfile
@@ -7,8 +7,8 @@ FROM python:${PYTHON_VERSION}-slim
 # the application crashes without emitting any logs due to buffering.
 ENV PYTHONUNBUFFERED=1
 
-# Define the program entrypoint file where your agent is started
-ARG PROGRAM_MAIN="src/agent.py"
+# Define the program entrypoint file where your agent is started.
+ARG PROGRAM_MAIN="{{.ProgramMain}}"
 
 # Create a non-privileged user that the app will run under.
 # See https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user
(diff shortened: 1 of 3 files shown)


pkg/agentfs/examples/python.pip.Dockerfile
```diff
@@ -48,5 +49,8 @@ RUN python "$PROGRAM_MAIN" download-files
 EXPOSE 8081
 
 # Run the application.
-CMD ["python", "$PROGRAM_MAIN", "start"]
+# The "start" command tells the worker to connect to LiveKit and begin waiting for jobs.
+# We use `exec` here to have more predictable signal handling when the container is deployed,
+# while still allowing us to expand shell variables.
+CMD ["sh", "-c", "exec python \"$PROGRAM_MAIN\" start"]
```
yoonsio: Was this tested? `sh -c` is going to be stripped anyways and we append `exec`, so not sure if it will work.
https://github.com/livekit/cloud-agents-secret-injector/blob/f956567e55a3fde32cb38961c5140ebdc1ce3377/main.go#L192

rektdeckard: Ah, I forgot about this. I only tested by building the images and running locally in docker.

rektdeckard: Then we either need to not use any variable here and fall back to a simpler means of overriding the entrypoint, or add some indirection where we rely on setting the `PROGRAM_MAIN` environment var, then reading from it in a script.

rektdeckard: OK @yoonsio I swapped this out with a simple string template, I think this will work a lot nicer in the future so long as we don't try to get fancy or have the goal of modifying people's existing dockerfiles.

PR #646, opened 2025-08-13
Make some more updates to the dockerfiles

bcherry:
Mostly just cleanup/consistency as I was writing docs

I tried to test these but I'm having trouble with both project detection and templating @rektdeckard  - not sure if there's something I'm doing wrong. I have tested the files in a project, I just can't get them to create from the CLI as expected.


bcherry:
@real-danm @rektdeckard I reopened this PR - would be a good update to get out especially with the healthcheck port removal

PR #649, opened 2025-08-17
check for empty client settings

real-danm:


yoonsio (approved):

cmd/lk/agent.go
```diff
@@ -478,7 +478,7 @@ func createAgentConfig(ctx context.Context, cmd *cli.Command) error {
 		if configExists && lkConfig.HasAgent() {
 			agentID = lkConfig.Agent.ID
 		} else {
-			agentID, err = selectAgent(ctx, cmd, false)
+			agentID, err = selectAgent(cmd, false)
```
rektdeckard: Is there a reason not to propagate ctx? this fn makes network requests, but i guess we're only ever cancelling it via SIGINT/SIGKILL

real-danm: good point on context, i updated to propogate it to the util.Await function

cmd/lk/agent.go
```diff
@@ -1186,6 +1191,10 @@ func getClientSettings(ctx context.Context, silent bool) (map[string]string, err
 		return nil, err
 	}
 
+	if clientSettingsResponse == nil {
```
yoonsio, line 1194: Can we have nil error and nil clientSettingsResponse at the same time?

rektdeckard: good point...i don't think twirp ever does that but i'm not positive

real-danm: probably not, but doesn't hurt to leave this in so it doesn't panic in that case

PR #655, opened 2025-08-19
Add generate dockerfile command

bcherry:
While testing #646 I made a couple updates that deserved their own PR

- [x] Add a `lk agent dockerfile` command to make new dockerfiles without making agents
- [x] Fix python entrypoint detection to ignore `__init__.py` files and also ensure agent.py/main.py are listed first
- [x] Add a little more output to the commands to tell you about the files being created and the project type detection used

image

### DIFF
--- a/cmd/lk/agent.go
+++ b/cmd/lk/agent.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"regexp"
 	"slices"
 	"strings"
@@ -110,6 +111,22 @@ var (
 					// we disable it entirely here and require multiple --secrets flags to be used.
 					DisableSliceFlagSeparator: true,
 					ArgsUsage:                 "[working-dir]",
+				},
+				{
+					Name:   "dockerfile",
+					Usage:  "Generate Dockerfile and .dockerignore for your project",
+					Before: createAgentClient,
+					Action: generateAgentDockerfile,
+					Flags: []cli.Flag{
+						silentFlag,
+						&cli.BoolFlag{
+							Name:     "overwrite",
+							Usage:    "Overwrite existing Dockerfile and/or .dockerignore if they exist",
+							Required: false,
+							Value:    false,
+						},
+					},
+					ArgsUsage: "[working-dir]",
 				},
 				{
 					Name:   "config",
@@ -378,6 +395,7 @@ func createAgent(ctx context.Context, cmd *cli.Command) error {
 	}
 
 	projectType, err := agentfs.DetectProjectType(workingDir)
+	fmt.Printf("Detected project type [%s]\n", util.Accented(string(projectType)))
 	if err != nil {
 		return fmt.Errorf("unable to determine project type: %w, please use a supported project type, or create your own Dockerfile in the current directory", err)
 	}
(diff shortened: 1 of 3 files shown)


cmd/lk/agent.go
```diff
@@ -378,6 +395,7 @@ func createAgent(ctx context.Context, cmd *cli.Command) error {
 	}
 
 	projectType, err := agentfs.DetectProjectType(workingDir)
+	fmt.Printf("Detected project type [%s]\n", util.Accented(string(projectType)))
```
rektdeckard, line 398: Let's just move this into `DetectProjectType()` so we don't lose it elsewhere

cmd/lk/agent.go
```diff
@@ -1213,3 +1231,68 @@ func requireConfig(workingDir, tomlFilename string) (bool, error) {
...
+		return fmt.Errorf("invalid working directory: %s", workingDir)
+	}
+
+	settingsMap, err := getClientSettings(ctx, cmd.Bool("silent"))
+	if err != nil {
+		return err
+	}
+
+	projectType, err := agentfs.DetectProjectType(workingDir)
+	fmt.Printf("Detected project type [%s]\n", util.Accented(string(projectType)))
```
rektdeckard, line 1250: and remove here

pkg/agentfs/docker.go
```diff
@@ -115,13 +122,16 @@ func validateEntrypoint(dir string, dockerfileContent []byte, dockerignoreConten
 				return err
 			}
 			if !d.IsDir() && strings.HasSuffix(d.Name(), projectType.FileExt()) {
+				// Exclude files like __init__.py which cannot be entrypoints
+				if d.Name() == "__init__.py" {
```
rektdeckard: Can probably add `__version__.py` too

rektdeckard (approved):
Few notes, otherwise looks good!

cmd/lk/agent.go
```diff
@@ -1213,3 +1231,68 @@ func requireConfig(workingDir, tomlFilename string) (bool, error) {
...
+	if err != nil {
+		return err
+	}
+
+	if writeDockerfile {
+		if err := os.WriteFile(dockerfilePath, dockerfileContent, 0644); err != nil {
+			return err
+		}
+
+		fmt.Printf("Wrote new %s\n", util.Accented("Dockerfile"))
```
rektdeckard, line 1287: "Created", since new file IMO

bcherry: well sometimes its an overwrite. do you still think "created" works?

pkg/agentfs/docker.go
bcherry: probably any __ file actually

cmd/lk/agent.go
bcherry: well i didn't really want it in the deploy command's output and also its annoying to add to detectProjectType because either you add it all over that function or you need to refactor it to a single return statement

real-danm (approved):

cmd/lk/agent.go
rektdeckard: That's a fair point. Fine as is, or "Generated"?

pkg/agentfs/docker.go
rektdeckard: I think `__main__.py` can be a valid entrypoint, in case you also want to use the folder as a module and an executable no? Not an expert.

cmd/lk/agent.go
bcherry: leaving it as-is

pkg/agentfs/docker.go
bcherry: yep good point - i excluded all others

bcherry:
@rektdeckard @real-danm do you know what the point of the `python_entrypoint` value in client settings is? it appears to just always be `main.py` but I don't think we need it at all anymore. can I ignore it? There seems to also be some outdated dockerfile stuff in there too - maybe the only thing we still need is the sdk min versions?

PR #657, opened 2025-08-20
Allow updating dispatch rule type.

dennwc:


cmd/lk/sip.go
```diff
@@ -277,6 +269,23 @@ var (
...
+								},
+								&cli.StringFlag{
+									Name:    "caller",
+									Aliases: []string{"individual"},
+									Usage:   "Sets a individual caller dispatch to a new room with a specific prefix",
+								},
+								&cli.StringFlag{
+									Name:  "callee",
+									Usage: "Sets a callee number dispatch to a new room with a specific prefix",
+								},
```
rektdeckard: FYI there are `MutuallyExclusiveFlags` if these options are incompatible with each other!

cmd/lk/sip.go
```diff
@@ -249,14 +249,6 @@ var (
 									Name:  "callee",
 									Usage: "Sets a callee number dispatch to a new room with a specific prefix",
 								},
-								&cli.BoolFlag{
-									Name:  "pin",
-									Usage: "PIN for a dispatch rule",
-								},
```
nishadmusthafa, line 255: Noticed this got removed. Was this intended ?

dennwc: Yes, it was unused it turns out. I might add it back later.

cmd/lk/sip.go
dennwc: TIL, thank you! Added the mutual exclusive group here.

PR #658, opened 2025-08-21
fix(agents): correctly parse and validate min sdk version

rektdeckard:


real-danm (approved):

rektdeckard:
@bcherry that's a great point, and not an easy one to deal with. We would have to clone the repo at the specified hash and then do project detection *in the repo*, which is a bigger can of worms, since they could use any build tools. Should we quietly allow this?

rektdeckard:
There's also the issue that the Dockerfile can literally do anything it wants after we check. It could install different package versions than specified in the lockfile or package file. I think we need a runtime solution.

davidzhao:
> Something we didn't discuss is what about git installs? We will see people with critical bugs in production who need to hotfix them ASAP and want to ship from git while the PR is merged and a new release is made. Whether from their own fork or from a branch on the main repo.
> 
> What will our checker do?

I added handling for git (as I had to test it directly from git before). it attributes it as `latest` after [my PR](https://github.com/livekit/livekit-cli/pull/648)

rektdeckard:
That works for the base, but you can totally specify a commit hash, branch, tag, etc. too

PR #659, opened 2025-08-22
feat(projects): start storing ID on project credentials

rektdeckard:
https://linear.app/livekit/issue/DPRO-461/save-project-id-in-cli-credentials

- Begins saving project ID on credentials when authenticating
- Adds ability to specify a specific project to auth with `lk cloud auth <project_id>`, which when backend support is added, will let us pre-select (and automatically confirm) project to authenticate.

### DIFF
--- a/cmd/lk/cloud.go
+++ b/cmd/lk/cloud.go
@@ -70,6 +70,11 @@ var (
 					Before: initAuth,
 					Action: handleAuth,
 					Flags: []cli.Flag{
+						&cli.StringFlag{
+							Name:     "id",
+							Usage:    "Project `ID` to authenticate. If not provided, you can choose from your existing projects in the browser",
+							Required: false,
+						},
 						&cli.BoolFlag{
 							Name:        "revoke",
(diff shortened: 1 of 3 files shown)


yoonsio (commented):
Do we query for projectID based on their current configs?

rektdeckard:
@yoonsio we don't backfill existing configured projects, just new auths from now on. If you want I can add a check and backfill if any project doesn't have an ID whenever config file is read?

cmd/lk/cloud.go
```diff
@@ -271,6 +276,7 @@ func tryAuthIfNeeded(ctx context.Context, cmd *cli.Command) error {
 	// get devicename
 	if err := huh.NewInput().
 		Title("What is the name of this device?").
+		Description("A short name you can use to find and manage generated API keys on the LiveKit Cloud dashboard").
```
bcherry, line 279: nice

bcherry (approved):

PR #661, opened 2025-08-22
support @next package version

bcherry:
this allows testing js 1.0 agents properly

i also have some provisional updates to the dockerfile - i'll keep this PR in draft until that's ready

### DIFF
--- a/pkg/agentfs/examples/node.Dockerfile
+++ b/pkg/agentfs/examples/node.Dockerfile
@@ -17,15 +17,11 @@ ENV PATH="$PNPM_HOME:$PATH"
 RUN apt-get update -qq && apt-get install --no-install-recommends -y ca-certificates && rm -rf /var/lib/apt/lists/*
 
 # Pin pnpm version for reproducible builds
-RUN npm install -g pnpm@9.15.9
+RUN npm install -g pnpm@10
 
 # Create a new directory for our application code
 # And set it as the working directory
 WORKDIR /app
-
-# Build stage
-# We use a multi-stage build to keep the runtime image minimal
-FROM base AS build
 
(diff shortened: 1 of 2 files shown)


pkg/agentfs/examples/node.Dockerfile
```diff
@@ -40,13 +36,15 @@ RUN pnpm install --frozen-lockfile
 COPY . .
 
 # Build the project
-RUN pnpm run build
+RUN pnpm exec tsc
```
lukasIO: what's the reason behind this change?

bcherry: trying to reduce the number of assumptions made. the previous version assumed your package.json has a script called `build` but that's not part of the default made with `pnpm init`. in the updated [quickstart](https://github.com/livekit/web/pull/1848/files#r2294820202) we are going to include various `pnpm pkg set "scripts.xxx"` commands but the fewer the better. so ensuring the default dockerfile works without a `build` script reduces the error surface area.

however - i haven't considered this thoroughly yet. of course it does include the assumption that you're using typescript and we don't currently have detection for non-TS projects. 

related, though, I do think we're going to start heavily pushing the [templated](https://github.com/livekit-examples/agent-starter-node/pull/4) quickstart very soon. so that makes things a little smoother but still doesn't totally absolve us of the need to support good dockerfile defaults that work for a range of projects.

bcherry: I'm switching this to assume the existence of build, download-files, and start scripts and just ensuring the quickstart sets those up and we have docs about it

rektdeckard (approved):

PR #664, opened 2025-08-25
feat: ability to set agent & metadata when creating token

davidzhao:


cmd/lk/token.go
```diff
@@ -107,6 +107,14 @@ var (
...
 							Usage: "Additional `VIDEO_GRANT` fields. It'll be merged with other arguments (JSON formatted)",
 						},
+						&cli.StringFlag{
+							Name:  "agent",
+							Usage: "Agent to dispatch to the room (identified by agent_name)",
+						},
+						&cli.StringFlag{
+							Name:  "dispatch-metadata",
+							Usage: "Dispatch metadata for the agent",
+						},
```
rektdeckard: Would `agent-metadata` be more intuitive? or is it very important to indicate this is only the agent metadata's starting state

davidzhao: this is actually different from the agent's metadata. it's a piece of metadata that is attached to the job.. I can rename it as the `job-metadata` instead.

https://github.com/livekit/protocol/blob/main/protobufs/livekit_agent_dispatch.proto#L36

rektdeckard: Ah, ok. What do people tend to use it for IRL? Leave it up to you with the name.

davidzhao: when you dispatch an agent, this provides information about the job to the agent.

PR #672, opened 2025-08-29
Fix broken python builds

bcherry:
I accidentally committed some work in https://github.com/livekit/livekit-cli/pull/661 that was not meant to be committed, but I didn't notice while rushing to get the 1.0 node support out.  that stuff broke python builds due to overwriting the server venv

### DIFF
--- a/pkg/agentfs/tar.go
+++ b/pkg/agentfs/tar.go
@@ -22,7 +22,6 @@ import (
 	"io"
 	"net/http"
 	"os"
-	pathpkg "path"
 	"path/filepath"
 	"strings"
 
@@ -44,6 +43,7 @@ var (
 	}
 
 	ignoreFilePatterns = []string{
+		".gitignore",
 		".dockerignore",
 	}
 )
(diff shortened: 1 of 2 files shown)


pkg/agentfs/tar.go
```diff
@@ -144,22 +133,20 @@ func UploadTarball(directory string, presignedUrl string, excludeFiles []string)
...
+				if strings.HasPrefix(relPath, exclude+"/") || strings.HasPrefix(relPath, exclude) {
 					logger.Debugw("excluding directory from tarball", "path", path)
 					return filepath.SkipDir
 				}
 			}
 
-			matched, err := pathpkg.Match(exclude, relPath)
+			matched, err := filepath.Match(exclude, relPath)
 			if err != nil {
 				return nil
```
rektdeckard, line 151: We do still need to write UNIX file paths to the tarball header. I made this change [here](https://github.com/livekit/livekit-cli/pull/666/files#diff-9a9462dcb33d39404e034294b66c3892d69cea945872a95d19041f8f6ad34be4R222) but it seems to have been removed.

bcherry: yeah ok i think i didn't get the revert right, i'll give it another stab. the goal is to be exactly what you did

rektdeckard (approved):

pkg/agentfs/tar.go
bcherry: upon further investigation we think the revert is correct

PR #673, opened 2025-08-30
fix dockerignore and tarball uploads for cloud agents

real-danm:
- loads the existing dockerignore if there is one
- uses the project type specific dockerignore if one isn't found
- add tests to ensure both top-level and subdirs do not include ignored files

these tests pass on my local (mac), will add a workflow to run tests on PRs/merges in a following PR

### DIFF
--- a/cmd/lk/agent.go
+++ b/cmd/lk/agent.go
@@ -432,7 +432,7 @@ func createAgent(ctx context.Context, cmd *cli.Command) error {
 		return err
 	}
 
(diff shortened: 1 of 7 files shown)


bcherry (approved):
Code looks solid and less brittle - do you have a windows machine available to test on?

rektdeckard (approved):

real-danm:
> Code looks solid and less brittle - do you have a windows machine available to test on?

i don't have a windows machine set up to test this. i added a test run to the GH actions including windows. these new changes pass although there is still some windows issue we need to fix, will address that in a follow-up pr

PR #677, opened 2025-09-09
Support file-mounted secret for cloud-agents

yoonsio:
* add `--secret-mount` flag to support file-mounted secrets.
* fix secret `overwrite` bug which required `overwrite` flag every time secrets are updated.

ref HA-278

### DIFF
--- a/cmd/lk/agent.go
+++ b/cmd/lk/agent.go
@@ -1055,15 +1055,16 @@ func listAgentSecrets(ctx context.Context, cmd *cli.Command) error {
 		return fmt.Errorf("unable to list agent secrets: %w", err)
 	}
 
+	// TODO (steveyoon): show secret.Kind.String() once cloud-agents is released
 	table := util.CreateTable().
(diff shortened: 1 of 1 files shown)


cmd/lk/agent.go
```diff
@@ -69,6 +70,12 @@ var (
 		Required: false,
 	}
 
+	secretsMountFlag = &cli.StringSliceFlag{
+		Name:     "secret-mount",
+		Usage:    "Local path to a secret file to be mounted on agent environment",
```
rektdeckard, line 74: Is it worth clarifying that (for now) this, and all other secrets, are only available at runtime?

yoonsio: Do you have any specific suggestion on what usage should read?

rektdeckard: I guess "on agent environment" is clear enough, on second thought.

rektdeckard: Rebase and your windows tests should pass 👍

PR #689, opened 2025-09-23
Adding CLI options to specify display name in CreateSipParticipant

alexlivekit:
@dennwc - Please note: I went with `--display-name "something something"` and `--no-display-name` which overrides it. Not sure if we like a better way there.


boks1971 (approved):
lgtm! cc: @rektdeckard for review as well

PR #694, opened 2025-09-30
fix(sip): implicitly paginate sip trunk queries to skirt twirp limits

rektdeckard:
https://linear.app/livekit/issue/CUS-276/livekit-cli-add-support-for-pagination-to-lk-sip-outboundinbound-list

Creates an `ExhaustivePaginatedList` helper function for chunking large data sets using `List*` SDK operations to skirt Twirp's 4MB payload limit. Implements it for `sip inbound list` and `sip outbound list`.

We should also apply this to other potentially large data sets for `room`, `dispatch`, `ingress` and `egress` list operations.

Worth noting: the SIP SDK does not seem to respect the `Limit` yet, but this change should be transparent and will resolve when SDKs are updated.


cmd/lk/proto.go
```diff
@@ -44,6 +45,16 @@ type protoTypeValidator[T any] interface {
 	Validate() error
 }
 
+type paginatedType[T any] interface {
+	protoType[T]
+	GetPage() *livekit.Pagination
```
nishadmusthafa, line 50: We plan to use this for phone numbers 
https://github.com/livekit/protocol/blob/main/protobufs/livekit_models.proto#L31
Should we account for this as well ?

cmd/lk/proto.go
```diff
@@ -99,6 +110,35 @@ func RequestDesc[T any, _ protoType[T]]() string {
...
+		if len(resultItems) > 0 {
+			iter(resultItems)
+		}
+		// List exhausted
+		if len(resultItems) < int(page.Limit) {
+			break
+		}
+	}
+	return nil
+}
```
nishadmusthafa, line 140: Do we need to worry about local memory if the api returns a large amount say thousands of phone numbers ?

rektdeckard: I don't know any modern machine that couldn't handle a few MB or even tens of MB or allocation, and this is a short-lived object. I would suggest we impose limit if and only if we get issues? This worked fine with 22K trunks in testing.

cmd/lk/proto.go
rektdeckard: Sure, I'll take a look

PR #695, opened 2025-09-30
support regional log pulls for cloud agents

real-danm:
requires: https://github.com/livekit/protocol/pull/1235

### DIFF
--- a/cmd/lk/agent.go
+++ b/cmd/lk/agent.go
@@ -463,7 +463,7 @@ func createAgent(ctx context.Context, cmd *cli.Command) error {
 			return err
 		} else if viewLogs {
 			fmt.Println("Tailing runtime logs...safe to exit at any time")
-			return agentfs.LogHelper(ctx, lkConfig.Agent.ID, "deploy", project)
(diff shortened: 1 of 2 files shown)


cmd/lk/replay.go
```diff
@@ -170,9 +170,9 @@ func loadReplay(ctx context.Context, cmd *cli.Command) error {
 	}
 
 	req := &replay.PlaybackRequest{
-		ReplayId:  cmd.String("id"),
-		RoomName:  cmd.String("room"),
-		StartTime: int64(cmd.Int("pts")),
+		ReplayId:     cmd.String("id"),
+		PlaybackRoom: cmd.String("room"),
+		SeekOffset:   int64(cmd.Int("pts")),
```
real-danm, line 185: cc @frostbyte73

pkg/agentfs/client.go
```diff
@@ -164,7 +165,7 @@ func (c *Client) uploadAndBuild(
 	return nil
 }
 
-func (c *Client) getAgentsURL() string {
+func (c *Client) getAgentsURL(prefix string) string {
```
yoonsio: Can we name it region instead of prefix to avoid confusion?

yoonsio (approved):

PR #700, opened 2025-10-02
Adding ability to specify headers on CreateSIPParticipant

alexlivekit:
Not sure we want to, but I've used this to pass headers to CreateSIPParticipant, which already supports the capability.

### DIFF
--- a/cmd/lk/sip.go
+++ b/cmd/lk/sip.go
@@ -370,8 +370,8 @@ var (
 									Value: 80 * time.Second,
 								},
 								&cli.StringSliceFlag{
-									Name:  "headers",
-									Usage: "Custom SIP headers in format 'Key:Value' (can be specified multiple times)",
+									Name:  "header",
+									Usage: "Custom SIP header in format 'Key:Value' (can be specified multiple times)",
 								},
 							},
 						},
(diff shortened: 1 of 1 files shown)


cmd/lk/sip.go
```diff
@@ -369,6 +369,10 @@ var (
 									Usage: "timeout for the call to dial",
 									Value: 80 * time.Second,
 								},
+								&cli.StringSliceFlag{
+									Name:  "headers",
```
dennwc:
```suggestion
									Name:  "header",
```
Because it is defined as an array, so one must repeat it: `--header K1:V1 --header K2:V2`.

cmd/lk/sip.go
```diff
@@ -1154,6 +1158,26 @@ func createSIPParticipant(ctx context.Context, cmd *cli.Command) error {
 		if cmd.Bool("wait") {
 			req.WaitUntilAnswered = true
 		}
+
+		// Parse headers from StringSliceFlag
+		if headers := cmd.StringSlice("headers"); len(headers) > 0 {
```
dennwc:
```suggestion
		if headers := cmd.StringSlice("header"); len(headers) > 0 {
```

dennwc (approved):

PR #702, opened 2025-10-08
handle post signed request

yoonsio:
Handle post presigned request with policy condition.
This is prerequisite for this feature and version enforcement is required.

ref HA-294

### DIFF
--- a/pkg/agentfs/tar.go
+++ b/pkg/agentfs/tar.go
@@ -28,7 +28,6 @@ import (
 	"strings"
 
 	"github.com/schollz/progressbar/v3"
-	"golang.org/x/sync/errgroup"
 
 	"github.com/livekit/livekit-cli/v2/pkg/util"
 	"github.com/livekit/protocol/livekit"
@@ -284,7 +283,7 @@ func UploadTarball(
 	)
 
 	if presignedPostRequest != nil {
-		if err := multipartUpload(presignedPostRequest.Url, presignedPostRequest.Values, &buffer, uploadProgress); err != nil {
+		if err := multipartUpload(presignedPostRequest.Url, presignedPostRequest.Values, &buffer); err != nil {
 			return fmt.Errorf("multipart upload failed: %w", err)
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


pkg/agentfs/tar.go
```diff
@@ -274,25 +283,79 @@ func UploadTarball(directory string, presignedUrl string, excludeFiles []string,
...
+		if err := multipartUpload(presignedPostRequest.Url, presignedPostRequest.Values, &buffer, uploadProgress); err != nil {
+			return fmt.Errorf("multipart upload failed: %w", err)
+		}
+	} else {
+		if err := upload(presignedUrl, &buffer, uploadProgress); err != nil {
+			return fmt.Errorf("upload failed: %w", err)
+		}
+	}
+
+	fmt.Println()
```
rektdeckard: Rogue newline? Or is this to clear the progressbar?

pkg/agentfs/tar.go
```diff
@@ -274,25 +283,79 @@ func UploadTarball(directory string, presignedUrl string, excludeFiles []string,
...
 		body, _ := io.ReadAll(resp.Body)
 		return fmt.Errorf("failed to upload tarball: %d: %s", resp.StatusCode, body)
 	}
-
-	fmt.Println()
 	return nil
 }
+
+func multipartUpload(presignedURL string, fields map[string]string, buf *bytes.Buffer, uploadProgress *progressbar.ProgressBar) error {
+	pr, pw := io.Pipe()
```
rektdeckard: Do you think we should include the `MaxSourceSize` on the signed post request so we can check if the buffer size would exceed exceed it first before trying to send a tarball we know in advance is too big? Can return a helpful error here about the size limit.

In practice, since we solved the problem of accidentally uploading `.venv`, have we seen anyone come close to this limit?

pkg/agentfs/tar.go
yoonsio: yeah I'll remove it

pkg/agentfs/tar.go
yoonsio: That is enforced from the server-side.

rektdeckard: Copy! The request is denied immediately based on the `content-length`?

PR #703, opened 2025-10-12
Adding the phone number apis

nishadmusthafa:


cmd/lk/phone_number.go
```diff
@@ -0,0 +1,443 @@
...
+	if err != nil {
+		return err
+	}
+
+	if cmd.Bool("json") {
+		util.PrintJSON(resp)
+		return nil
+	}
+
+	return listAndPrint(ctx, cmd, func(ctx context.Context, req *livekit.SearchPhoneNumbersRequest) (*livekit.SearchPhoneNumbersResponse, error) {
```
davidzhao: nit: this is a bit hard to read. any chance to break it out?

cmd/lk/phone_number.go
```diff
@@ -0,0 +1,443 @@
...
+	"github.com/livekit/livekit-cli/v2/pkg/util"
+	"github.com/livekit/protocol/livekit"
+	lksdk "github.com/livekit/server-sdk-go/v2"
+	"github.com/urfave/cli/v3"
+)
+
+var (
+	PhoneNumberCommands = []*cli.Command{
+		{
+			Name:   "phonenumber",
```
davidzhao: should we just use `number` for short? wdyt?

nishadmusthafa: yeah, that works well for the cli. Want to call out that the apis use phonenumber though.

cmd/lk/phone_number.go
```diff
@@ -0,0 +1,452 @@
...
+						},
+						&cli.IntFlag{
+							Name:  "limit",
+							Usage: "Maximum number of results (default: 50)",
+							Value: 50,
+						},
+						&cli.StringFlag{
+							Name:  "page-token",
+							Usage: "Token for pagination (empty for first page)",
+						},
```
rektdeckard: I think this is the best solution for now, but we were discussing either running implicit (behind the scenes) pagination (example here: https://github.com/livekit/livekit-cli/pull/694) or doing a TUI pager that would abstract away having to find and pass an explicit `page-token`.

As it stands, will I need to format my results as `--json` to see the page token in order to be able to fetch the next page?

rektdeckard: I mention it here because we can be pretty sure that at least a few of our customers will have hundreds of numbers.

cmd/lk/phone_number.go
```diff
@@ -0,0 +1,452 @@
...
+					Name:   "get",
+					Usage:  "Get a phone number from a project",
+					Action: getPhoneNumber,
+					Flags: []cli.Flag{
+						&cli.StringFlag{
+							Name:  "id",
+							Usage: "Use phone number ID for direct lookup",
+						},
+						&cli.StringFlag{
+							Name:  "phonenumber",
```
rektdeckard: Nit: `number` is simpler

cmd/lk/phone_number.go
```diff
@@ -0,0 +1,452 @@
...
+						jsonFlag,
+					},
+				},
+				{
+					Name:   "purchase",
+					Usage:  "Purchase phone numbers from inventory",
+					Action: purchasePhoneNumbers,
+					Flags: []cli.Flag{
+						&cli.StringSliceFlag{
+							Name:     "phonenumbers",
```
rektdeckard: Nit: `numbers` is nicer, and how we refer to them in other SIP contexts

cmd/lk/phone_number.go
```diff
@@ -0,0 +1,452 @@
...
+					Name:   "release",
+					Usage:  "Release phone numbers",
+					Action: releasePhoneNumbers,
+					Flags: []cli.Flag{
+						&cli.StringSliceFlag{
+							Name:  "ids",
+							Usage: "Use phone number IDs for direct lookup",
+						},
+						&cli.StringSliceFlag{
+							Name:  "phonenumbers",
```
rektdeckard: Ditto

cmd/lk/phone_number.go
nishadmusthafa: So I haven't implemented pagination fully in the backend.  I'm going to remove this param for now, fix it in the backend and then create another PR for pagination. 

on the approach for pagination, for the implicit pagination, would we run into memory issues if we load large lists ? I guess I'd like to understand concerns against the page-token approach.

PR #706, opened 2025-10-15
Refactor shared lib to use `fs.FS`

yoonsio:
Refactor shared lib to use `fs.FS` instead of `workingDir` to support both real and virtual file systems.
Unlike previous behavior, all symlink files are excluded.

ref HA-304

### DIFF
--- a/pkg/agentfs/tar.go
+++ b/pkg/agentfs/tar.go
@@ -165,9 +165,7 @@ func UploadTarball(
 	tarWriter := tar.NewWriter(gzipWriter)
 	defer tarWriter.Close()
 
-	root := "."
-
-	err = fs.WalkDir(directory, root, func(path string, d fs.DirEntry, err error) error {
+	err = fs.WalkDir(directory, ".", func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


pkg/agentfs/utils.go
kuba--: This can be removed - it's not used, and with current fs.FS it needs to be refactored

pkg/agentfs/tar.go
```diff
@@ -166,62 +165,30 @@ func UploadTarball(
...
-			}
-
-			// Copy file contents directly without progress bar
-			_, err = io.Copy(tarWriter, file)
-			if err != nil {
-				return fmt.Errorf("failed to copy file content for %s: %w", path, err)
-			}
-			return nil
-		}
-
```
rektdeckard, line 217: I think there was at least one user who needed symlinking to support a monorepo with multiple agent projects and shared code...but I guess until we have a better solution for that use-case it is okay to remove this.

yoonsio: Yeah hopefully in those cases, the original files are part of the source so this won't be an issue.

rektdeckard: They had shared code and each agent at the same level, so they would have to make some changes. They were doing this:
```
/shared
/agent-a
  agent.py
  pyptoject.toml
/agent-b
  agent.py
  pyptoject.toml
```

But still think its okay to break. Let's do a minor version bump when we release this.

PR #713, opened 2025-10-23
support agent region

real-danm:
added a select themed with the standard form to select from available regions.

### DIFF
--- a/cmd/lk/agent.go
+++ b/cmd/lk/agent.go
@@ -140,7 +140,9 @@ var (
 							},
 						}},
 					}},
-					Flags:                     []cli.Flag{},
+					Flags: []cli.Flag{
+						regionFlag,
+					},
 					ArgsUsage:                 "[AGENT-NAME]",
 					DisableSliceFlagSeparator: true,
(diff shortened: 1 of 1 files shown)


cmd/lk/agent.go
```diff
@@ -554,7 +553,31 @@ func createAgent(ctx context.Context, cmd *cli.Command) error {
...
+				Options(huh.NewOptions(regionOptions...)...).
+				Value(&region).
+				WithTheme(util.Theme).
+				Run(); err != nil {
+				return err
+			}
+		} else {
+			// we shouldn't ever get here, but if we do, just default to us-east
+			logger.Debugw("no available regions found, defaulting to us-east. please contact LiveKit support if this is unexpected.")
+			region = "us-east"
```
yoonsio, line 578: Shouldn't default region be empty?

real-danm: hmm i wasn't sure here. if it's empty then whichever region picks it up based on dns of the api call would deploy it. not sure if that is better or to use us-east

PR #715, opened 2025-10-27
Display owner for cloud-agent version

yoonsio:
ref HA-289

### DIFF
--- a/cmd/lk/agent.go
+++ b/cmd/lk/agent.go
@@ -959,17 +959,22 @@ func listAgentVersions(ctx context.Context, cmd *cli.Command) error {
 	}
 
 	table := util.CreateTable().
-		Headers("Version", "Current", "Status", "Created At", "Deployed At")
+		Headers("Version", "Current", "Status", "Owner", "Created At", "Deployed At")
 
 	// Sort versions by created date descending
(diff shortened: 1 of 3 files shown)


cmd/lk/agent.go
```diff
@@ -959,17 +959,22 @@ func listAgentVersions(ctx context.Context, cmd *cli.Command) error {
...
 	table := util.CreateTable().
-		Headers("Version", "Current", "Status", "Created At", "Deployed At")
+		Headers("Version", "Current", "Status", "Owner", "Created At", "Deployed At")
 
 	// Sort versions by created date descending
 	slices.SortFunc(versions.Versions, func(a, b *lkproto.AgentVersion) int {
 		return b.CreatedAt.AsTime().Compare(a.CreatedAt.AsTime())
 	})
 	for _, version := range versions.Versions {
+		owner := version.Owner
```
rektdeckard, line 969: This is an owner ID, not a name, correct? Unsure if that's really useful info in the context of the CLI. Maybe we should just leave it to the dashboard to surface the full name of the version owner?

yoonsio: Yeah does livekit-cli has access to user table?

rektdeckard: No, and we don't have session-token auth so we can't call cloud-api methods directory very easily.

Issue #712, opened 2025-10-22
the lk agent delete doesn't have silent mode


mhamri:
I am trying to delete the agent during cicd and recreate it again. but `lk agent delete --id $ExistingAgent` always is asking for confirmation and there isn't any `--silent` mode in it

real-danm:
for cicd you can use this repo to make things simpler: https://github.com/livekit/deploy-action
I have not yet implemented delete there yet though, so added the silent flag for delete in https://github.com/livekit/livekit-cli/pull/723

PR #723, opened 2025-11-05
add a silent mode for agent delete

real-danm:
requested in https://github.com/livekit/livekit-cli/issues/712


rektdeckard (approved):
Note to self:
- Support `silent` mode for all commands
- Informational output should go to `stderr` and command results should go to `stdout` so you could always do `2>/dev/null` for a poor man's silent mode, while still getting the result of an operation when needed

PR #726, opened 2025-11-07
Changing up how dispatch rules are assigned to numbers

nishadmusthafa:


cmd/lk/phone_number.go
```diff
@@ -161,6 +161,36 @@ func createPhoneNumberClient(ctx context.Context, cmd *cli.Command) (*lksdk.Phon
...
+func getPhoneNumberToDispatchRulesMap(ctx context.Context, cmd *cli.Command) (map[string][]string, error) {
+	_, err := requireProject(ctx, cmd)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get project: %w", err)
+	}
+
+	sipClient := lksdk.NewSIPClient(project.URL, project.APIKey, project.APISecret, withDefaultClientOpts(project)...)
+
+	// List all dispatch rules
+	resp, err := sipClient.ListSIPDispatchRule(ctx, &livekit.ListSIPDispatchRuleRequest{})
```
rektdeckard, line 175: Just noting, we still haven't resolved the issue with max twirp message size, either here or in SDK. Probably very few customers hitting that but we need to come back to the idea of auto-paginating results when we do things like this.

nishadmusthafa: this is similar to what we've done here -> https://github.com/livekit/livekit-cli/pull/694/files ? I can take that up after landing this.

rektdeckard: Yeah, that was the idea for a short-term fix. What we probably want to for real tho is something like this if the results are requested in JSON format (complete results are expected) but have real pagination in the CLI for when you expect a table format (just call with `nextPageToken` on arrow key or something).

PR #734, opened 2025-12-03
Incorporating changes in the list phone number api

nishadmusthafa:


cmd/lk/phone_number.go
```diff
@@ -370,66 +295,41 @@ func listPhoneNumbers(ctx context.Context, cmd *cli.Command) error {
...
-	phoneNumberToRules := dispatchRulesRes.rules
-	if dispatchRulesRes.err != nil {
-		// Log but don't fail
-		if cmd.Bool("verbose") {
-			fmt.Fprintf(cmd.ErrWriter, "Warning: failed to get dispatch rules: %v\n", dispatchRulesRes.err)
-		}
-		phoneNumberToRules = make(map[string][]string)
+	fmt.Printf("Total phone numbers: %d", resp.TotalCount)
+	if resp.OfflineCount > 0 {
+		fmt.Printf(" (%d offline)", resp.OfflineCount)
```
rektdeckard, line 310: I assume these are total counts from cloud-io, not just the counts for the page we are viewing?

cmd/lk/phone_number.go
rektdeckard, line 345: Will this status now take into account the "offlineness" of the number, or is it still active for all numbers that aren't pending or released?

cmd/lk/phone_number.go
nishadmusthafa: Yupp this is across all numbers and not just the page.

cmd/lk/phone_number.go
nishadmusthafa: Yes. This does show offline if there is not routing configured for the number and will show active if it works. That logic is being handled in cloud-io now.

PR #739, opened 2025-12-18
[wip] add lock file validation

andrewnitu:
if you do `lk agent create` before doing `uv sync` (or the equivalent action in other languages), it will start deploying but fail the build partway through due to missing uv.lock. The agent will also be stuck in a bad state.

This PR adds a check to make sure the relevant lock file is there before attempting to deploy.

### DIFF
--- a/pkg/agentfs/detect.go
+++ b/pkg/agentfs/detect.go
@@ -115,3 +115,40 @@ func DetectProjectType(dir fs.FS) (ProjectType, error) {
 
 	return ProjectTypeUnknown, errors.New("project type could not be identified; expected package.json, requirements.txt, pyproject.toml, or lock files")
 }
+
+// ValidateLockFiles checks if the required lock files exist for the given project type.
+// This ensures that dependency sync commands (e.g., uv sync, npm install) have been run
+// before deployment, preventing build failures.
+func ValidateLockFiles(dir fs.FS, projectType ProjectType) error {
+	switch projectType {
+	case ProjectTypePythonUV:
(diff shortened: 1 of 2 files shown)


pkg/agentfs/detect.go
```diff
@@ -115,3 +115,40 @@ func DetectProjectType(dir fs.FS) (ProjectType, error) {
 
 	return ProjectTypeUnknown, errors.New("project type could not be identified; expected package.json, requirements.txt, pyproject.toml, or lock files")
 }
+
+// ValidateLockFiles checks if the required lock files exist for the given project type.
+// This ensures that dependency sync commands (e.g., uv sync, npm install) have been run
+// before deployment, preventing build failures.
+func ValidateLockFiles(dir fs.FS, projectType ProjectType) error {
```
andrewnitu, line 122: this function definitely need to be double-checked. It's probably correct for uv projects but I'm not familiar with all the other kinds of project.

yoonsio (approved):
nice!

pkg/agentfs/detect.go
rektdeckard: One thing we could do is allow this check to be bypassed when `--skip-sdk-check` is set, as we already do with similar project detection tasks in case somebody is using a non-standard package manager or something.

rektdeckard: Otherwise it looks correct to me. You could add some tests similar to this: https://github.com/livekit/livekit-cli/blob/main/pkg/agentfs/sdk_version_check_test.go#L9

PR #742, opened 2025-12-30
prompt to tail logs upon `lk agent deploy`

andrewnitu:
It now prompts for logs:
image

and also shows the logs (from the new replica):
image


real-danm (approved):
lgtm, would check with @yoonsio too in case he knew of a reason not to show these after a deploy

PR #767, opened 2026-02-12
Add CLI flag to parse user timestamps from h264 SEI NALs

chenosaurus:
adds a `--publish-frame-metadata` flag that enables the feature in the go server sdk to parse H264/H265 stream for timestamps embedded in SEI NALs.


cmd/lk/room.go
```diff
@@ -177,6 +177,10 @@ var (
 							Usage: "Format to use when reading H.264 from file or socket, \"annex-b\" OR \"length-prefixed\"",
 							Value: "annex-b",
 						},
+						&cli.BoolFlag{
+							Name:  "publish-user-timestamp",
+							Usage: "When publishing H.264/H.265, attach the LKTS user timestamp trailer to each encoded frame (timestamp is sourced from preceding H.264 SEI user_data_unregistered when present)",
```
boks1971: mix of H.264/H.265 mention here. Mentions only H.264 SEI in details. Would be good to add H.265 details also.

Also, in join.go flag ☝️ , it only mentions H.264.

Also, `join.go` commands are deprecated. Would be good to not add new ones there.

boks1971 (approved):

cmd/lk/join.go
```diff
@@ -57,10 +57,15 @@ var (
...
 					TakesFile: true,
-					Usage: "`FILES` to publish as tracks to room (supports .h264, .ivf, .ogg). " +
+					Usage: "`FILES` to publish as tracks to room (supports .h264, .h265, .ivf, .ogg). " +
 						"can be used multiple times to publish multiple files. " +
 						"can publish from Unix or TCP socket using the format '<codec>://<socket_name>' or '<codec>://<host:address>' respectively. Valid codecs are \"h264\", \"h265\", \"vp8\", \"opus\"",
 				},
+				&cli.BoolFlag{
+					Name:   "attach-frame-metadata",
+					Usage:  "Parse H264/H265 SEI for LKTS frame metadata (user timestamp and frame ID) and re-attach the packet trailer to each encoded frame",
+					Hidden: true,
```
boks1971, line 67: do you want to leave this `hidden`?

boks1971: I think this command itself is deprecated. So, may be fine.

cmd/lk/room.go
```diff
@@ -177,6 +177,10 @@ var (
 							Usage: "Format to use when reading H.264 from file or socket, \"annex-b\" OR \"length-prefixed\"",
 							Value: "annex-b",
 						},
+						&cli.BoolFlag{
+							Name:  "attach-frame-metadata",
+							Usage: "When publishing H.264/H.265, parse SEI user_data_unregistered for LKTS frame metadata (user timestamp and frame ID) and re-attach the packet trailer to each encoded frame",
```
boks1971: The comment here vs in join.go looks different, intentional? The other location mentions SEI.

PR #810, opened 2026-04-02
add console, simulate, and agent dev commands

theomonnom:


cmd/lk/console.go
```diff
@@ -0,0 +1,288 @@
...
+		},
+		&cli.StringFlag{
+			Name:  "entrypoint",
+			Usage: "Agent entrypoint `FILE` (default: auto-detect)",
+		},
+	},
+	Action: runConsole,
+}
+
+func runConsole(ctx context.Context, cmd *cli.Command) error {
```
rektdeckard, line 84: I'm seeing weird behavior when switching between audio and text modes in the console:
- In text mode, messages are printed out-of-order (agent responses appearing before user)
- Switching to text mode mid-call does not seem to work; responses continue to arrive in audio modality and text responses don't get printed

cmd/lk/agent_run.go
```diff
@@ -0,0 +1,287 @@
...
+	&cli.StringFlag{
+		Name:    "api-key",
+		Usage:   "LiveKit API `KEY`",
+		Sources: cli.EnvVars("LIVEKIT_API_KEY"),
+	},
+	&cli.StringFlag{
+		Name:    "api-secret",
+		Usage:   "LiveKit API `SECRET`",
+		Sources: cli.EnvVars("LIVEKIT_API_SECRET"),
+	},
```
rektdeckard: These flags are already available globally; no need to redefine or even declare them on this command. But FYI, the precedence of resolution is:
1. project specified in `livekit.toml` (actually this is probably too high, should probably be after flags but separate issue)
2. explicit flags (`--api-key`, `--api-secret`, `--url`)
3. explicit project (`--project`)
4. environment variables
5. default project

README.md
```diff
@@ -54,6 +54,30 @@ git clone https://github.com/livekit/livekit-cli && cd livekit-cli
 make install
 ```
 
+### Building with console support
```
rektdeckard: let's maybe comment this out until we feel the feature is stable

Formula/lk.rb
```diff
@@ -0,0 +1,47 @@
+# typed: false
+# frozen_string_literal: true
+
+# This formula is meant for a custom Homebrew tap (e.g. livekit/homebrew-livekit).
+# It installs a prebuilt binary with console support (PortAudio + WebRTC AEC).
+# Usage: brew install livekit/livekit/lk
+class Lk < Formula
```
rektdeckard: Are we sure we are unable to distribute prebuilt binaries through homebrew's core package repository?

rektdeckard (approved):
Approving for expediency, and I am in fact able to build and test (mostly) successfully against agents on `theo/1.6.0` branch -- but there are some rough edges particularly around console project/entrypoint detection, text mode, and ergonomics that need some attention. I have not had success testing simulations, as I assume there is still outstanding backend work for that?

rektdeckard:
@claude review

Formula/lk.rb
davidzhao: yeah homebrew has to build it from source.. and sign it with apple distribution keys

theomonnom:
For the text mode issue, I'll handle that in a separate PR. This requires some protocol changes inside our new RemoteSession

PR #816, opened 2026-04-07
cloud-agents environment support

yoonsio:
Adds `--env` flag to `create`, `destroy` and `update-secrets` commands.
* create/deploy: environment defaults to `production`.
* update-secrets: can associate secrets with multiple environments.
* secrets: show environments column in agent secrets table.

ref AP-638


go.mod
```diff
@@ -208,3 +210,5 @@ require (
 	mvdan.cc/sh/v3 v3.12.0 // indirect
 	sigs.k8s.io/json v0.0.0-20241014173422-cfa47c3a1cc8 // indirect
 )
+
+replace github.com/livekit/server-sdk-go/v2 => /Users/syoon/src/server-sdk-go
```
real-danm, line 214: saw your note, just another reminder

go.mod
```diff
@@ -34,6 +34,8 @@ require (
 	k8s.io/apimachinery v0.34.1
 )
 
+replace github.com/livekit/protocol => /Users/syoon/src/protocol
```
real-danm, line 37: same here

real-danm (approved):
lgtm, will the block a release until server side is ready? have to coordinate the rollout if so

cmd/lk/agent.go
```diff
@@ -97,6 +97,12 @@ var (
 		Required: false,
 	}
 
+	envFlag = &cli.StringSliceFlag{
+		Name:     "env",
+		Usage:    "Deployment environment(s). For create/deploy, specifies the target environment (defaults to 'production'). For update-secrets, assigns environment(s) to the secret. Can be specified multiple times (e.g. --env staging --env production).",
+		Required: false,
+	}
```
shawnyang-coder, line 104: Why only create/deploy/update-secrets? What about other rpc like status/update? 

Can an agent (identified by unique agent_id) be in both Staging and Prod ? or only one env at any time?

yoonsio: status/update shows agent details including all environments. Yes, multiple environments (prod, staging) can exists within a single agent (agent_id).

PR #820, opened 2026-04-09
log region selection in agent creation

andrewnitu:
adds this log line at the bottom.
image

previously if you selected a region and you forgot / not paying attention you'd have no idea where it was deploying to :-)


cmd/lk/agent.go
```diff
@@ -598,6 +598,7 @@ func createAgent(ctx context.Context, cmd *cli.Command) error {
 				Run(); err != nil {
 				return err
 			}
+			fmt.Printf("Using region [%s]\n", util.Accented(region))
```
rektdeckard: Small nit, I've been meaning to update all informational prints to go to `stderr` so as not to get mixed in with real outputs when piping or redirecting. Can you update this?

```suggestion
			fmt.Fprintf(os.Stderr, "Using region [%s]\n", util.Accented(region))
```

rektdeckard (approved):

andrewnitu:
@rektdeckard committed your suggestion and tried it out. It does go to stderr now.

PR #821, opened 2026-04-10
feat: support web console opener, deprecate sandbox creation

rektdeckard:
- Adds support for Web Console opener in `lk token create`, `lk room join`
- `lk agent init` no longer defaults to creating a sandbox (and the flag is deprecated and hidden)
- `lk agent init` and `lk app create` now print Web Console link on completion
- `lk agent init` and `lk app create` now prompt to auto-install dependencies


yoonsio (approved):
Do we still support template projects?

rektdeckard:
@yoonsio yes, the templates are unchanged except that the agent name is now used to print a link to the web-based console instead of a sandbox (for python and node agent templates):

Screenshot 2026-04-09 at 11 02 01 PM

real-danm (approved):

cmd/lk/app.go
```diff
@@ -374,10 +375,20 @@ func setupTemplate(ctx context.Context, cmd *cli.Command) error {
...
+			WithTheme(util.Theme).
+			Run(); err != nil {
+			return err
+		}
+	}
 	if install {
 		fmt.Println("Installing template...")
 		if err := doInstall(ctx, bootstrap.TaskInstall, appName, verbose); err != nil {
-			return err
+			fmt.Fprintf(os.Stderr, "Warning: installation failed: %v\n", err)
```
Topherhindman, line 394: Was the intent here to make install failures non fatal? I’m asking because it looks like we continue into `post_create` and `cleanup` even if `doInstall` fails, and I wasn’t sure if that could leave someone with a partially initialized app that’s harder to deal with?

rektdeckard: Yes, but the issue we're averting is that the user does not have required installation tools (`uv` or `pnpm`). It's totally fine to leave a user in a project that hasn't had dependencies installed yet. What's *not* fine is to fail early and not complete the setup process which involves de-gitting and removing the template file from the repo.

PR #831, opened 2026-04-22
clean up byoc - run image handling before project type check

real-danm:
was from a bad merge on my part, just need the --image handling to run before the project detect code since BYOC won't have a local package manager file


andrewnitu (approved):
it was like this when i originally reviewed it haha

PR #835, opened 2026-04-27
Hosted-agents:  first-cut agentName changes

shawnyang-coder:
- Add -name xxx flag to lk agent init/create/register/update, with create/deploy being required. 
- In init, agentName can optionally read from toml file, or the first command arg (even without --name flag). 
- Upon receiving responses, write agentName back to the toml file.


cmd/lk/agent.go
```diff
@@ -176,6 +185,7 @@ var (
 					Action: createAgent,
 					Before: createAgentClient,
 					Flags: []cli.Flag{
+						agentNameFlag(true),
```
shawnyang-coder, line 188: For Create and Deploy, I made the flag required. The intention is to force the migration.

go.mod
```diff
@@ -37,6 +37,8 @@ require (
 	k8s.io/apimachinery v0.34.1
 )
 
+replace github.com/livekit/server-sdk-go/v2 => ../server-sdk-go
```
shawnyang-coder, line 40: I need this line to build locally since this change spans multiple repos. How do I merge the PRs so that the two repos do not break each other during the interim? @yoonsio @real-danm

shawnyang-coder: I think we will need to merge the server-sdk-go PR first. Cut a release. Then update the go.mod here to point to that release right? But what about those livekit-cli used in the field?

yoonsio: Yes that is correct

PR #838, opened 2026-05-08
Update SDK to support new SIP options.

dennwc:


hechen-eng (approved):

cmd/lk/sip.go
```diff
@@ -480,6 +476,105 @@ func listSetFlag(cmd *cli.Command, setName string) ([]string, bool) {
...
+	if vals, ok := listSetFlag(cmd, "codecs"); ok {
+		codecs, err := parseSIPCodecs(vals)
+		if err != nil {
+			return nil, err
+		}
+		m := getMedia()
+		m.Codecs = codecs
+		if val, ok := optBoolFlag(cmd, "no-default-codecs"); ok {
+			m.OnlyListedCodecs = val
+		}
```
rektdeckard, line 574: Nit: I personally feel like the indirection here doesn't make this simpler.

```suggestion
		noDefaultCodecsFlag := "no-default-codecs"
		if cmd.isSet(noDefaultCodecsFlag) {
			m.OnlyListedCodecs = cmd.Bool(noDefaultCodecsFlag)
		}
```

rektdeckard (approved):

cmd/lk/sip.go
dennwc: It doesn't require an extra var with the flag name, though.

PR #844, opened 2026-05-17
add support for packet trailer features

haffi96:


haffi96:
@chenosaurus some notes

1. Changes in agent.go were required to successfully build 🤔 but not required for or related to this feature
2. `--publish-user-timestamp` enables attaching both user timestamp and frame id

PR #846, opened 2026-05-21
feat(lk): env command preserves existing file content unless --overwrite

rektdeckard:
https://linear.app/livekit/issue/GRW-547/improve-lk-app-env-write-ergonomics

- `lk app env --write` now preserves existing entries in the target file (upserting only the vars we touch)
- This behavior can be overridden with `--overwrite` 

Sorry, this was an old ticket that got lost 😅

### DIFF
--- a/pkg/bootstrap/bootstrap.go
+++ b/pkg/bootstrap/bootstrap.go
@@ -318,9 +318,11 @@ func ReadDotEnv(rootDir string, filePath string) (map[string]string, error) {
 
 // WriteDotEnv writes envMap to filePath under rootDir. When overwrite is false
 // and the file already exists, envMap is merged into the file in place: keys
-// in envMap are updated, keys not in envMap are preserved along with comments
-// and blank lines, and any envMap keys not yet in the file are appended. When
-// overwrite is true (or the file does not exist), the file is written fresh.
+// in envMap are updated (preserving any inline comments and the surrounding
+// whitespace on those lines), keys not in envMap are preserved verbatim along
+// with comments and blank lines, and any envMap keys not yet in the file are
+// appended. When overwrite is true (or the file does not exist), the file is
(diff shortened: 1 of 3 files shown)


pkg/bootstrap/bootstrap.go
```diff
@@ -303,15 +304,93 @@ func PrintDotEnv(envMap map[string]string) error {
...
 	if err != nil {
 		return err
 	}
-	envLocalPath := path.Join(rootDir, filePath)
 	return os.WriteFile(envLocalPath, []byte(envContents+"\n"), 0700)
 }
 
+var envAssignmentRe = regexp.MustCompile(`^(\s*(?:export\s+)?)([A-Za-z_][A-Za-z0-9_]*)\s*=`)
+
+func mergeDotEnv(existing string, envMap map[string]string) (string, error) {
```
Topherhindman, line 404: can we add some tests around this merge behavior?

Topherhindman (commented):
this seems fine, however, i feel like dotenv files have enough weird cases that it'd be good for us to have some coverage for.  some that could be good:
- preserves comments/blank lines
- updates existing keys
- appends missing keys 
- keeps `export`
- handles inline comments
- handles missing destination file
- verifies `overwrite=true` still fully replaces

pkg/bootstrap/bootstrap.go
rektdeckard: done!

pkg/portaudio/pa_src
Topherhindman, line 1: is this change intentional?

Topherhindman (approved):

pkg/portaudio/pa_src
rektdeckard: It is unintentional but permissible 😆

PR #850, opened 2026-05-28
feat(lk): project resolution refactor, other UX

rektdeckard:
- Makes project resolution more deterministic
- Adds `Printer` struct to make it easier to do file redirection, respect `--quiet` and `--verbose` flags globally, and to keep output from getting mixed in with debug information. Status messages and command outputs should not be printed manually with `fmt` functions from now on, and should instead use e.g. `out.Status()`, `out.Result()`, etc.
- Improves UX of confirmations to make it harder to confuse yes/no:

BEFORE:
CleanShot 2026-05-28 at 00 43 36@2x
AFTER:
CleanShot 2026-05-28 at 00 43 54@2x


cmd/lk/cloud.go
```diff
@@ -284,7 +284,7 @@ func tryAuthIfNeeded(ctx context.Context, cmd *cli.Command) error {
 	}
 
 	// poll for keys
-	fmt.Printf("Please confirm access by visiting:\n\n   %s\n\n", authURL.String())
+	out.Statusf("Please confirm access by visiting:\n\n   %s\n", authURL.String())
```
Topherhindman, line 287: `out.Statusf` is `--quiet`-suppressed, so `lk cloud auth -q` i think would hide the verification URL, right?

pkg/util/printer_test.go
```diff
@@ -0,0 +1,64 @@
...
+func TestPrinter_QuietSuppressesOnlyStatus(t *testing.T) {
+	var out, err bytes.Buffer
+	p := NewPrinter(&out, &err, true)
+
+	p.Status("breadcrumb")
+	p.Statusf("formatted %s", "breadcrumb")
+	p.Warnf("warn %d", 1)
+	p.Resultf("result\n")
+
+	assert.Empty(t, "", err.String()[:0], "sanity")
```
Topherhindman: is this actually testing anything?

cmd/lk/agent.go
```diff
@@ -567,13 +568,13 @@ func createAgent(ctx context.Context, cmd *cli.Command) error {
...
 		if !silent {
-			fmt.Printf("Using agent configuration [%s]\n", util.Accented(tomlFilename))
+			out.Statusf("Using agent configuration [%s]", util.Accented(tomlFilename))
 		}
 	} else {
 		lkConfig = config.NewLiveKitTOML(subdomainMatches[1]).WithDefaultAgent()
 	}
 	if !silent {
-		fmt.Printf("Creating new agent deployment\n")
+		out.Status("Creating new agent deployment")
```
Topherhindman: `--silent` and `--quiet` overlap here. these `out.Status*` calls already respect `--quiet` internally, but they're also wrapped in `if !silent`, so they're double gated, where as [here](https://github.com/livekit/livekit-cli/pull/850/changes/37dcffc612b0d317f0a6991396008275013b1c84#diff-4a70574ec74b4fc8888fd29a16e49065fec8bd1833148322247a5522810a1c89R1509) its only gated by `--quiet`.  

should we fold `--silent`'s status suppression into `--quiet` and drop the `if !silent` wrappers?

Topherhindman (approved):
couple nits, but overall definitely a good improvement!

cmd/lk/agent.go
rektdeckard: Yes, good call. I'll just collapse these into one flag (`--quiet`). The cool thing about this is we almost don't need it at all, cuz now that everything happens on the expected FD you can do `cmd > /dev/null 2>&1` to mimic silent, or `cmd 2>/dev/null` to suppress statuses.

rektdeckard: Still nice to have the flag for it tho

cmd/lk/cloud.go
rektdeckard: Correct, it doesn't print the URL -- but it does still open a browser window automatically if currently on an interactive terminal.

cmd/lk/agent.go
rektdeckard: Done. `--silent` is still respected, but it is an alias for `--quiet` and now all this logic is handled within the Printer rather than done manually.

pkg/util/printer_test.go
rektdeckard: I mean yeah it is a dumb check that can't ever fail. I thank Claude for this gem.

PR #851, opened 2026-05-28
Add pagination support to egress list

biglittlebigben:


cmd/lk/egress.go
```diff
@@ -587,14 +591,32 @@ func listEgress(ctx context.Context, cmd *cli.Command) error {
...
+			}
+
+			res, err = egressClient.ListEgress(ctx, req)
+			if err != nil {
+				return err
+			}
+
+			resItems := res.Items
+			if remaining := limit - len(items); limit > 0 && len(resItems) > remaining {
+				resItems = resItems[:remaining]
```
milos-lk: just to double check since pages get older as we paginate, if the limit cuts mid-page - should we return its tail rather than the head?

milos-lk: I forgot if we order asc or desc within a page - checking it

milos-lk: ingress does newest last within a page - I guess that will apply to egress as well so this might be valid.

milos-lk:
nit: The pagination tests cover the RPC control flow well — request count, page tokens, and when pagination stops. They don’t yet assert the user-visible behavior of --limit: which egress IDs are returned and in what order. (might be related to my other comment above)

milos-lk (approved):

PR #853, opened 2026-06-01
chore(build): rework build process

rektdeckard:
- Removes the `console` build tag, including console feature by default on all builds on all platforms
- Updates build and release workflows to support new constraints around cross-compilation:
  - Linux and Windows binaries are built on an ubuntu runner, with platform-specific artifacts (ALSA, MinGW) compiled via zig cc 
  - MacOS binaries are built on a native mac runner, using cgo and native libraries (CoreAudio) 
- Streamlines `Makefile`, moving logic into go directives so that build is simplified enough to be run with `go build cmd/lk`, but build process is still discoverable by CodeQL 
- Updates `fish_autocomplete`, which had not been generated in quite some time, and validates it is current in CI now
- Some code cleanup to appease linter


davidzhao (approved):

theomonnom (approved):
lgtm! thanks!

PR #858, opened 2026-06-03
Remove trailing slash when building url

s3bw:
This is host dependent, but providing `example.com/?url...` can be treated as a different path to `example.com?url...`.

Instead we rely on what ever the user provides as `base_url` and append the query params to that base url.


davidzhao:
I think we should use a proper URL builder here, instead of string manipulations. the current code seems fragile and a bit difficult to reason with.

i.e. 

```go
u, err := url.Parse(baseUrl)
if err != nil {
    log.Fatal(err)
}
q := u.Query()
q.Set("url", ...)
u.RawQuery = q.Encode()
fmt.Println(u)
```

PR #866, opened 2026-06-09
agent: launch agents via python -m livekit.agents (framework-owned CLI)

theomonnom:


theomonnom:
Folded the launcher change into #863.

PR #867, opened 2026-06-09
simulate(tui): in-TUI scenario save (no overwrite) + copy-to-clipboard; drop export-scenarios subcommand

theomonnom:


cmd/lk/simulate_ci.go
```diff
@@ -55,6 +56,11 @@ func runSimulateCI(ctx context.Context, config *simulateConfig) error {
 				fmt.Fprintf(os.Stderr, "Agent logs: %s\n", agent.LogPath)
 			}
 		}
+		if config.mode == modeGenerateFromSource && run != nil {
+			if path, err := writeGeneratedScenariosTemp(run); err == nil && path != "" {
+				fmt.Fprintf(os.Stderr, "Generated scenarios: %s\n", path)
```
rektdeckard: All print statements should use the new `Printer` interface. There is a global `out` instance of it with helpers for debug messages (`out.Status(...)`) and command output (`out.Result(...)`). This helps us manage output, respect the `--quiet` flag, and keep random stuff from cluttering stdout.

rektdeckard (approved):

cmd/lk/simulate_ci.go
theomonnom: Done: all simulate prints now go through `out` — breadcrumbs via `Status`/`Statusf` (silenced by `--quiet`), results and the gh `::error::` annotations via `Resultf`, the broken-agent note on `out.Err` (warning-class, never quieted). For the writer-based streaming (simLog, agent-startup log forwarding) I added `Printer.StatusWriter()`, which returns `io.Discard` under `--quiet`.

PR #861, opened 2026-06-04
feat(session): support Node.js agents in the local session daemon

toubatbrian:
> [!IMPORTANT]
> **Depends on / must wait for: livekit/agents-js#1706** — *feat(console): console CLI runner + AgentsConsole session wiring (text mode)*.
>
> The Go daemon spawns the JS agent with `node <entry> console --connect-addr <addr>`. The JS `console --connect-addr` TCP runner that accepts this is added in agents-js#1706, which is **not yet merged**. The underlying `TcpSessionTransport` (agents-js#1693) is already merged, but without #1706 there is no JS console TCP entrypoint to dial the daemon. **Do not merge this until agents-js#1706 lands.**

## Summary

Extends the `lk agent session` local daemon to drive **Node.js/TypeScript** agents, not just Python. Previously Node projects were rejected at project-detection time; this wires up the Node spawn path so a JS/TS agent is driven over the exact same text-mode console protocol (`console --connect-addr <addr>`) as Python.

## Changes

- **`cmd/lk/agent_utils.go`** — `detectProject` no longer gates on Python-only. It now also accepts Node projects (already recognized by `DetectProjectRoot` via `package.json`) and the error message reflects both supported runtimes.
- **`cmd/lk/simulate_subprocess.go`** — extracted interpreter/argv resolution into `buildAgentCommand`, branching on project type:
  - Python: `<python> <entry> <args>` (uv prefixes `run python`)
  - Node: `node [--experimental-strip-types] <entry> <args>`
  - Added `findNodeBinary` (resolves `node` from `PATH`) and `isTypeScriptEntry` so a `.ts`/`.mts`/`.cts` entrypoint runs directly via Node's type-stripping loader — no build step required.
- **`cmd/lk/session_daemon.go`** — dropped the stale `TODO(node)`; the daemon now spawns either runtime via `buildConsoleArgs`.

## Underlying spawned command

```
node --experimental-strip-types <entry> console --connect-addr 127.0.0.1:<port>
```

(`--experimental-strip-types` is added only for TypeScript entrypoints.)

## Test plan

Verified end-to-end against a JS agent (`@livekit/agents`, with agents-js#1706 checked out locally):

- [x] `lk agent session start examples/src/console_text_agent.ts` — detects the Node project, spawns the JS agent, connects to the Go TCP console server, completes the text-mode handshake (`Session started.`)
- [x] `lk agent session say "What is the weather in Tokyo?"` — round-trips a turn including a `getWeather` function-tool call and the agent reply
[...]

PR #868, opened 2026-06-09
feat(agent): Node.js agent support for lk agent dev/start/console

u9g:
Stacked on #861. Extends Node.js agent support from the session daemon to the full `lk agent dev` / `start` / `console` flow.

## Changes

**Node agents in dev/start/console** (9128955)
- `dev`/`start` spawn the agents-js `dev` subcommand for Node projects (Python keeps `start --dev`)
- Log levels normalized per runtime: agents-js accepts only lowercase, Python expects uppercase
- `--reload-addr` is Python-only and no longer passed to Node agents
- Entrypoint candidate lists per runtime (`agent.ts`/`agent.js` + `src/` fallbacks), with cwd-relative resolution so monorepo example dirs work
- `resolveCredentials` honors `IsSet("url")` so an explicit `--url` wins

**Env file loading for Node agents** (2430a24)
- Node agents spawned by dev/start/console inherited only the shell environment, so credentials in a project `.env` were never seen (Python agents conventionally call `load_dotenv()` themselves; Node ones don't)
- A known env file discovered in the project dir is now passed to Node's built-in dotenv via `--env-file` (Node ≥ 20.6; skipped silently on older Nodes). Shell-exported variables still take precedence over file values

**Fail fast on pre-connect job crashes** (7660aa2)
- A job that crashes before the console TCP connection (e.g. missing API key in the entrypoint) leaves the worker process alive, so `lk agent console` previously sat on the spinner for the full 60s connect timeout
- Agent output is now scanned for crash markers (Python's `"job crashed"` shutdown reason, agents-js's FATAL `console mode failed:`) and the connect wait aborts immediately with the recent logs; the session daemon does the same through its ready pipe

## Testing
- Unit tests for `buildAgentCommand` (Node + Python argv), env-file discovery, and the fail-signal watcher
- E2E against agents-js (`examples/src/basic_agent.ts`, TS entrypoint via type stripping + `--env-file`) and Python agents @ a61578853 (`voice_agents/basic_agent.py`): full voice console session in both
- Crash repro: removing the `.env` now fails in ~4s with the traceback instead of timing out at 60s


cmd/lk/simulate_subprocess.go
```diff
@@ -148,23 +220,34 @@ type AgentStartConfig struct {
...
+		major, minor, version, ok := nodeVersion(cfg.Dir, nodeBin)
+		args := make([]string, 0, len(cfg.CLIArgs)+3)
 		if isTypeScriptEntry(cfg.Entrypoint) {
+			if !nodeVersionAtLeast(major, minor, ok, 22, 6) {
+				return "", nil, fmt.Errorf("running a TypeScript entrypoint directly requires Node >= 22.6 (found %s); upgrade Node or point at built JS output", version)
+			}
 			args = append(args, "--experimental-strip-types")
 		}
+		// --env-file requires Node >= 20.6; skip the flag (old behavior) on older Nodes.
+		if envFile := agentfs.FindEnvFile(cfg.Dir); envFile != "" && nodeVersionAtLeast(major, minor, ok, 20, 6) {
```
theomonnom: @rektdeckard What's your thoughts on auto-loading the env file?

rektdeckard: Realistically, agent code should already be doing this itself, no? And if it isn't (user has hardcoded keys or whatever) that's assumed to be intentional?

u9g: true, I added it for ease-of-use for running the examples, but for real users, more than likely they have already implemented the key flow themselves. It might just make more sense to support passing additional node args via an environment variable.

rektdeckard: ++, probably allowing arbitrary arguments to be forwarded is smart. Seen it with the `--` separator, e.g. 
```
lk agent console src/main.py -- --log-level=DEBUG
```

u9g: done. we now forward `-- --env-file=.env`. `--log-level` is a bad example for this because we explicitly set that within livekit-cli, and this feature's goal is to allow users to set their own node/python interpreter flags.

also, Brian's PR which this PR is based on adds `--experimental-strip-types`, which I just checked and is no longer required since node 24 (current LTS). Is it worth adding the flag unconditionally as we do now, or given that we now provide a way to pass the flag manually, should we no longer pass it ourselves?

toubatbrian: We can drop it, I'll fix this in my branch

rektdeckard: Maybe anything we have explicit command line parameters for that just normalize and forward to the agent, including all the stuff that maps debug levels and such, can be dropped. We lean into "pass args yourself".

theomonnom: +1 for being explicit, let's just forward everything to agents

theomonnom:
For framework specific "magic strings", let's try to move all of them at one place.

https://github.com/livekit/livekit-cli/blob/deb98ff87cbb04a8d95fad22258c52ff419370a2/cmd/lk/simulate_health.go#L29

theomonnom (approved):
Awesome!! Node finally has a console mode!

cmd/lk/agent_run_test.go
theomonnom, line 1: Lot of tests aren't super useful here

u9g: I just cleaned up the tests.

toubatbrian (approved):

PR #869, opened 2026-06-09
agent CLI: friendly errors for old SDK version + worker start failures

theomonnom:


theomonnom:
Folded into #867 instead of a separate PR.

PR #870, opened 2026-06-10
simulate: --concurrency flag + launch the agent with --simulation

theomonnom:
## What

- `lk agent simulate --concurrency N`: sets the new `SimulationRun.Create.Request.concurrency` — max simulate jobs running in parallel for the run. The server clamps it to the project's concurrent-simulate quota; omitted keeps the server default (5).
- The agent subprocess is now launched with the framework's new `--simulation` flag, which disables the worker load limit so the run can saturate the agent instead of getting `WS_FULL`-throttled.

## Dependencies

- livekit/protocol#1613 (pinned to the PR branch commit — re-pin after merge)
- livekit/agents#6036 (the `--simulation` flag; older livekit-agents won't recognize it, so `thinCLIMinVersion` should be bumped to the release that includes it before this ships)
- Related: livekit/agents-private#68, livekit/cloud-api-server#1892

Stacked on #867, hence the base branch.

### DIFF
--- a/cmd/lk/simulate.go
+++ b/cmd/lk/simulate.go
@@ -73,6 +73,10 @@ var simulateCommand = &cli.Command{
 			Aliases: []string{"n"},
 			Usage:   "Number of scenarios to generate",
 		},
+		&cli.IntFlag{
+			Name:  "concurrency",
+			Usage: "Max simulations running in parallel (default: server-side limit)",
+		},
 		&cli.StringFlag{
 			Name:  "scenarios",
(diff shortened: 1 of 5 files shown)


theomonnom:
Added in this PR: when a run completes without a summary, the CLI now shows the server-provided reason (`run.error`, persisted by livekit/agents-private#70 — e.g. *"can't summarize: user data recording (observability) is disabled for this project..."*) instead of the generic "summary not available" message, in both TUI and CI output.

PR #873, opened 2026-06-11
simulate: strip ANSI escapes from the agent log file

theomonnom:
The agent subprocess logs in colored format, and `lk-simulate-*.log` captured the raw lines — so the file users open after a failure was littered with ANSI escape sequences. The file write now strips them (same `ansiEscapeRe` already used for room extraction); in-memory lines and the TUI stream are unchanged.


theomonnom:
Folded into #867 instead.

PR #876, opened 2026-06-15
agent_simulation: --audio runs simulations over the agent's media pipeline

Bobronium:
Passes SimulationMode on the Create request; default stays TEXT.

### DIFF
--- a/cmd/lk/simulate.go
+++ b/cmd/lk/simulate.go
@@ -81,7 +81,7 @@ var simulateCommand = &cli.Command{
 		},
 		&cli.BoolFlag{
 			Name:  "audio",
-			Usage: "Simulate over audio (TTS + the agent's full media pipeline) instead of text",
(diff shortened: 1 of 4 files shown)


cmd/lk/simulate.go
```diff
@@ -79,6 +79,10 @@ var simulateCommand = &cli.Command{
 			Name:  "scenarios",
 			Usage: "Path to a scenarios `FILE` (yaml). If omitted, scenarios are generated from the agent's source",
 		},
+		&cli.BoolFlag{
+			Name:  "audio",
+			Usage: "Simulate over audio (TTS + the agent's full media pipeline) instead of text",
```
```suggestion
			Usage: "Simulate speech-to-speech interactions using the agent's full audio pipeline. By default, simulations run in text-only mode.",
```

theomonnom (approved):

PR #880, opened 2026-06-17
fix(install): verify release checksum before extracting in install-cli.sh

### LINKED ISSUE #647: CLI install script should validate checksum
# Request
[install script](https://github.com/livekit/livekit-cli/blob/main/install-cli.sh) should validate the checksum for security

mcdgavin:
## Summary

`install-cli.sh` piped the downloaded archive straight into `tar` with no integrity
check. This downloads the archive and the release's `checksums.txt`, verifies the
archive's SHA-256 against the published checksum (failing closed on a missing or
mismatched checksum), then extracts.

## Scope / honest caveat

`checksums.txt` is fetched from the same release over HTTPS, so this guards against
corrupted/partial downloads and accidental mismatches — it is **not** a substitute for
signature verification against an out-of-band key. The stronger follow-up would be to
sign `checksums.txt` (e.g. cosign/sigstore via goreleaser) and verify the signature
here.
I'm happy to do that separately if you'd like to enable signing.

## Test

Verified against the v2.16.6 release: the checksum parse + match succeeds on the real
archive, and a tampered archive is correctly rejected. `bash -n` passes. (The script is
Linux-only — it aborts on macOS — so it relies on `sha256sum`, with an added presence check.)

Fixes #647

Thanks to @xtreme-sameer-vohra for raising it.

### DIFF
--- a/install-cli.sh
+++ b/install-cli.sh
@@ -34,6 +34,7 @@ abort() { printf "%s\n" "$@" >&2; exit 1; }
 [ -n "${BASH_VERSION:-}" ] || abort "This script requires bash"
 [ -d "$INSTALL_PATH" ]     || abort "Could not install, $INSTALL_PATH doesn't exist"
 command -v curl >/dev/null || abort "cURL is required and is not found"
+command -v sha256sum >/dev/null || abort "sha256sum is required and is not found"
 
 OS="$(uname)"
 case "$OS" in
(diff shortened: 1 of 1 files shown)


davidzhao (approved):
lgtm!

davidzhao:
Thanks for the fix Gavin!

mcdgavin:
Thanks for your review @davidzhao. Any interest in exploring release signing? I created a draft pr here (#903) to address the signing, but would need a maintainer to test the actual signing with a pre-release tag push.

davidzhao:
@mcdgavin thanks for the suggestion. we'll review this one

PR #884, opened 2026-06-24
fix(agent init): surface dependency install failures clearly + don't show install command if the cli installed deps

u9g:
## Summary

Improves how `lk agent init` (and `lk app create`) handles the dependency-install step.

Previously a failed install (e.g. missing Node/pnpm) was downgraded to a single-line warning and the command still exited 0, so it was easy to miss — buried between status lines and the template's "agent created" banner.

This branch:

1. **Surfaces install failures prominently.** `doInstall` now captures the install task's stderr and reports the underlying command's own error (e.g. `env: node: No such file or directory`) instead of go-task's generic `exit status N` wrapper. The failure is rendered as a bold header with a red bar prefix on each error line and a yellow bar on the follow-up guidance. The agent is still created (install is best-effort).

2. **Signals a successful install to `post_create`.** On success, sets `LIVEKIT_DEPS_INSTALLED=1` so a template's `post_create` task can skip the now-redundant install hint (e.g. `pnpm install`) via a `status:` guard — mirroring how `LIVEKIT_AGENT_NAME`/`LIVEKIT_SANDBOX_ID` already gate template output. No-op until the template opts in (companion PR on `agent-starter-node`).

## Before / after (failure)

```
# before
Warning: installation failed: task: Failed to run task "install": exit status 127   (exit 0)

# after
⚠  Installation failed — dependencies were NOT installed
┃ env: node: No such file or directory
┃ Fix your toolchain, then re-run the install step manually in ./my-agent.
```

## Testing

- Verified the failure path renders the real command error.
- Verified `LIVEKIT_DEPS_INSTALLED=1` makes a `status:`-guarded template skip the install hint on success and keep it when install isn't run.

### DIFF
--- a/cmd/lk/app.go
+++ b/cmd/lk/app.go
@@ -457,6 +457,10 @@ func setupTemplate(ctx context.Context, cmd *cli.Command) error {
 			}
 			fmt.Fprintf(&b, "%sFix your toolchain, then re-run the install step manually in ./%s.", fixPrefix, appName)
 			out.Warnf("%s", b.String())
+		} else {
+			// Signal a successful install to post_create so the template can skip
+			// printing the now-redundant install hint (guarded via `status:`).
+			os.Setenv("LIVEKIT_DEPS_INSTALLED", "1")
 		}
 	}
 	if err := doPostCreate(ctx, cmd, appName, verbose); err != nil {
--- a/cmd/lk/app.go
+++ b/cmd/lk/app.go
@@ -15,15 +15,18 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"maps"
 	"os"
 	"regexp"
 	"strings"
 
 	"github.com/charmbracelet/huh"
+	"github.com/charmbracelet/lipgloss"
 	"github.com/urfave/cli/v3"
 
 	"github.com/livekit/livekit-cli/v2/pkg/bootstrap"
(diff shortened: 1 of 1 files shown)


cmd/lk/app.go
```diff
@@ -438,7 +441,26 @@ func setupTemplate(ctx context.Context, cmd *cli.Command) error {
...
+			// prominent warning. Each line of the underlying command's error gets
+			// a red bar prefix so the actual failure stands out from the guidance.
+			header := lipgloss.NewStyle().Foreground(util.Warning()).Bold(true).
+				Render("⚠  Installation failed — dependencies were NOT installed")
+			errPrefix := lipgloss.NewStyle().Foreground(util.Error()).Render("┃ ")
+			fixPrefix := lipgloss.NewStyle().Foreground(util.Warning()).Render("┃ ")
+			var b strings.Builder
+			b.WriteString(header + "\n")
+			for _, line := range strings.Split(strings.TrimRight(err.Error(), "\n"), "\n") {
+				b.WriteString(errPrefix + line + "\n")
```
rektdeckard: You're gonna get a warning about string concatenation here. Just do multiple calls to `b.WriteString`, it's more performant (not that it matters really).

cmd/lk/app.go
```diff
@@ -438,7 +441,26 @@ func setupTemplate(ctx context.Context, cmd *cli.Command) error {
...
+			}
+			fmt.Fprintf(&b, "%sFix your toolchain, then re-run the install step manually in ./%s.", fixPrefix, appName)
+			out.Warnf("%s", b.String())
```
rektdeckard: Warnf is already a formatter, you can do this in one go.

rektdeckard (approved):
Looks good. Nonblocking nits.

PR #886, opened 2026-06-24
deployment support

yoonsio:
Deployment support for cloud-agents and dispatch

ref AP-638


cmd/lk/agent.go
```diff
@@ -805,8 +816,14 @@ func deployAgent(ctx context.Context, cmd *cli.Command) error {
 		}
 	}
 
+	agentDeployment := ""
+	if cmd.IsSet("deployment") {
+		agentDeployment = cmd.String("deployment")
+		fmt.Printf("Using deployment [%s]\n", util.Accented(agentDeployment))
```
rektdeckard: please use the new printer interface for all command output (`out.Statusf`, `out.Result`, etc.)

cmd/lk/token.go
```diff
@@ -474,6 +481,7 @@ func createToken(ctx context.Context, c *cli.Command) error {
 				return err
 			}
 		case string(util.OpenTargetConsole):
+			// TODO (steveyoon): pass deployment to console when supported
```
rektdeckard: You can make this change now, just modify `ConsoleURLParams` to include the deployment. I will make sure we handle it in web, but no downside to just passing it already.

cmd/lk/agent.go
```diff
@@ -1660,3 +1746,10 @@ func generateAgentDockerfile(ctx context.Context, cmd *cli.Command) error {
 
 	return nil
 }
+
+func formatTime(t time.Time) string {
```
rektdeckard: Can you check if any other commands that print times could use this? Should put it in `utils` or something and make sure we format time consistently.

rektdeckard (approved):
LGTM, just some small nits.

cmd/lk/agent.go
```diff
@@ -244,6 +244,17 @@ var (
 					DisableSliceFlagSeparator: true,
 					ArgsUsage:                 "[working-dir]",
 				},
+				{
+					Name:   "promote",
```
JackNDwyer, line 264: @rektdeckard do you agree with this implementation?

cmd/lk/token.go
rektdeckard: actually @JackNDwyer already did it

JackNDwyer: Oh yes, did telephony dispatch and console already . On prod behind FF

PR #891, opened 2026-06-27
display console link after agent start

davidzhao:
also fixes agent dev assuming it's localhost

the agent sends its project subdomain and agent-name, so we can reliably send to the right project/agent in console


rektdeckard:
@davidzhao is there a corresponding agents branch I can test this on?

EDIT: found it https://github.com/livekit/agents/pull/6260

cmd/lk/agent_run.go
```diff
@@ -188,6 +285,23 @@ func runAgentDev(ctx context.Context, cmd *cli.Command) error {
...
+	// agent in the browser. Printed once, even across hot reloads (link stays valid).
+	var consoleLinkOnce sync.Once
+	cfg.OnServerInfo = func(agentName, wsURL string) {
+		if link := cloudConsoleURL(wsURL, agentName); link != "" {
+			consoleLinkOnce.Do(func() {
+				// Delay briefly so the link prints after the agent's own startup
+				// logs rather than getting buried in them.
+				time.AfterFunc(time.Second, func() {
+					out.Status("")
+					out.Statusf("Agent console: %s", link)
```
rektdeckard: Tiny nit, I would format the link in a bright color (`util.Accented`). Nice to have would also be to wrap with [OSC 8](https://terminfo.dev/extensions/osc-8-hyperlinks) to make it a clickable link in supporting terminals (there are many).

rektdeckard (approved):
LGTM, and working as expected with agent feature branch.

cmd/lk/agent_run.go
rektdeckard: We do this in the bootstrap task: https://github.com/livekit-examples/agent-starter-python/blob/main/taskfile.yaml#L61

theomonnom (approved):

cmd/lk/agent_run.go
davidzhao: done!

PR #892, opened 2026-06-29
simulate: allow pointing simulation to running agent via `--agent-name`

Bobronium:
Add `--agent-name` argument to `lk agent simulate`:

> Run against an already-running agent instead of spawning one locally. Pass the registered `NAME` or "" to target the project's default agent (the one that auto-joins every room). Requires --scenarios.


theomonnom (approved):
lgtm, is the CI failure related to this PR?

cmd/lk/simulate.go
```diff
@@ -251,34 +256,56 @@ func runSimulate(ctx context.Context, cmd *cli.Command) error {
...
+		projectDir  string
+		projectType agentfs.ProjectType
+		entrypoint  string
+		liveAgent   bool
+		err         error
+	)
+
+	// --agent-name (even empty) means: run against an already-running agent,
+	// don't spawn one. An empty name targets the project's default agent — the
+	// one that auto-joins every room (registered with no agent_name).
```
```suggestion
	// --agent-name (even empty) means: run against an already-running agent,
	// don't spawn one. https://docs.livekit.io/agents/server/agent-dispatch/#automatic
```

Bobronium:
Yup, CI failure seem to be unrelated.

PR #894, opened 2026-06-30
fix(agent): trust the resolved livekit-agents version in the SDK check; add Windows test runner

u9g:
The Python agent SDK-version check (`lk agent start/dev/console`, and the build-time check) could reject a project whose installed `livekit-agents` actually satisfies the minimum. This PR makes the check trust the resolved/installed version, and adds Windows test coverage for the package.

## 1. Parse the resolved version from `uv.lock` (`pkg/agentfs`)

`checkUvLock` matched a `livekit-agents = "x"` line that never appears in `uv.lock` — the file uses poetry.lock's TOML layout (`[[package]]` / `name` / `version`). So the resolved version was never found and the check fell back to the `pyproject.toml` constraint, misreading the lower bound (`>=1.0` → `1.0`) and failing it against the minimum (e.g. 1.6.0). Now it uses the same `[[package]]` regex as `checkPoetryLock`, so the resolved version is read and preferred.

## 2. Resolve the *installed* version for the run check (`cmd/lk`)

`lk agent start/dev/console` now resolves the installed `livekit-agents` via the project interpreter (`importlib.metadata`), the way the Node path already does — accurate regardless of installer (uv/pip/poetry) and not fooled by a loose version constraint. For uv it reads the existing environment with `--no-sync`, so the pre-flight check never syncs or downloads. When dependencies aren't installed (e.g. before a sync), it falls back to static project-file parsing (the uv.lock fix above, then constraints).

## 3. Windows CI (`.github/workflows/windows.yaml`)

`test.yaml` covers Linux/macOS Go tests and `session-e2e.yaml`'s `e2e` job covers Linux/macOS session e2e, but Windows ran essentially no Go tests: mingw GCC can't compile the WebRTC APM C++ (MSVC SEH), and the native cgo link overflows the command-line limit. This adds a single `windows.yaml` that — like `.goreleaser.yaml` — cross-compiles on Linux with zig and runs natively on `windows-latest`:

- One `cross-build` job produces `lk.exe` plus a test binary per package (discovered via `go list`, currently `cmd/lk`, `pkg/agentfs`, `pkg/bootstrap`, `pkg/util`).
- `windows-tests` (needs `cross-build`) runs every package's tests from its directory so cwd-relative fixtures resolve; runs on PRs, no secrets.
- `windows-session-e2e` (needs `cross-build`) drives the session lifecycle against the prebuilt `lk.exe`; skipped on PRs since it needs LiveKit secrets.

Both consumers depend on the one `cross-build`, so the expensive cgo compile happens once. The Windows jobs were moved out of `session-e2e.yaml` (now Linux/macOS only) into this workflow. Windows is amd64-only here, matching the prior session-e2e Windows arm; `build.yaml` still builds windows/arm64 for release.

## Tests

- `pkg/agentfs`: real `uv.lock` fixture (replacing a fake one that only matched the broken regex) + a regression test that a loose pyproject floor doesn't override a satisfying resolved version.
- `cmd/lk`: `resolvePythonAgentVersion` / `checkPythonSDKVersion` exercised against a real uv project with a local stub `livekit-agents` package (known version, no network, no real SDK) — reads installed version, rejects too-old, installed beats a loose constraint, and falls back to file parsing when not installed. These run on Linux, macOS, and Windows.

## Note

This is the first time the full Go suite runs on Windows, so the initial run may surface pre-existing Windows-specific issues in tests unrelated to this change.


Bobronium:
Should we perhaps just check the actual installed version instead?

This would save us from parsing any requirements/lock files and guarantee that this is the version that will be used:

```bash
uv run python -c \
'from importlib.metadata import version; \
print(version("livekit-agents"))'
```
```
1.6.3
```

cmd/lk/simulate_subprocess.go
```diff
@@ -180,6 +180,53 @@ func checkNodeSDKVersion(cfg AgentStartConfig) error {
 	return nil
 }
 
+// pythonResolveVersionScript prints the installed livekit-agents version, or
+// exits non-zero if it isn't importable.
+const pythonResolveVersionScript = `import importlib.metadata as m; print(m.version("livekit-agents"))`
```
theomonnom: Let's make sure this isn't executing any livekit-agents code. Otherwise, this could be a security vulnerability.

Let's make sure importlib is only reading metadata

u9g:
> Should we perhaps just check the actual installed version instead?
> 
> This would save us from parsing any requirements/lock files and guarantee that this is the version that will be used:
> 
> ```shell
> uv run python -c \
> 'from importlib.metadata import version; \
> print(version("livekit-agents"))'
> ```
> 
> ```
> 1.6.3
> ```

The cloud build (for create and deploy subcommands on the lk cli) resolves dependencies from the project/lock files, so uv.lock is the source of truth for what will actually run in deployment; the locally installed version could differ (or not exist) and would be the wrong thing to check.

cmd/lk/simulate_subprocess.go
u9g: Well, the interpreter starts, and the interpreter runs all .pth files of all dependencies which means all those files will be ran, but this is just a check before running `uv run`, so I don't see what the vulnerability could be here.

Bobronium:
> The cloud build (for create and deploy subcommands on the lk cli) resolves dependencies from the project/lock files, so uv.lock is the source of truth for what will actually run in deployment; the locally installed version could differ (or not exist) and would be the wrong thing to check.

Totally fair, I've misinterpreted the scope initially.
 
> For uv it reads the existing environment with --no-sync, so the pre-flight check never syncs or downloads.

I'd argue we should just add `--no-sync` to `findPythonBinary` to make it deterministic and consistent with the case when UV is not found and bare Python is used.

u9g:
> > The cloud build (for create and deploy subcommands on the lk cli) resolves dependencies from the project/lock files, so uv.lock is the source of truth for what will actually run in deployment; the locally installed version could differ (or not exist) and would be the wrong thing to check.
> 
> Totally fair, I've misinterpreted the scope initially.
> 
> > For uv it reads the existing environment with --no-sync, so the pre-flight check never syncs or downloads.
> 
> I'd argue we should just add `--no-sync` to `findPythonBinary` to make it deterministic and consistent with the case when UV is not found and bare Python is used.

This is a great point, I did it.

Bobronium (approved):
LGTM!

PR #898, opened 2026-07-09
fix(simulate): keep the job list visible when the summary overflows short terminals

u9g:
## Problem

`lk agent simulate --view <run>` in a short terminal (e.g. 31 rows) rendered the full run summary below the job list with no height budget. The view exceeded the screen, the header and top of the job list scrolled off, and a 10-job run showed only jobs 7–10 at all times.

## Fix

- **Clamp the summary in the list view** to the height remaining under the job list (accounting for the logs pane, toast, and hint bar), ending with a dim `↓ N more summary lines · press S to expand` marker. The view now always fits the terminal, so the whole job list stays visible.
- **`S` toggles an expanded summary** rendered in place of the job list, windowed and scrollable (arrows/PgUp/PgDn/mouse wheel), collapsed with `S`/`esc`/`q` — the same pattern as the job-detail and description views. Capital `S` avoids the existing `s` (save scenarios) binding.
- The `S` binding and its hint-bar entry appear only when the summary was actually clamped (it stays available to collapse an expanded view).

## Testing

New `simulate_tui_layout_test.go` reproduces the report: a 31×156 terminal with 10 jobs and a long summary must fit in 31 rows with the first and last scenarios visible; expanded mode fits, scrolls, and clamps scroll offsets; short summaries render unclamped with no `S` hint. Full `go test ./cmd/lk/` passes.

### DIFF
--- a/cmd/lk/simulate_tui.go
+++ b/cmd/lk/simulate_tui.go
@@ -176,6 +176,13 @@ type simulateModel struct {
 	logPinnedTotal  int
 	showDescription bool
 	descScrollOff   int
+	// summaryExpanded shows the run summary full-height (scrollable) instead
+	// of the job list; in the list view the summary is clamped to fit.
+	// summaryTruncated records whether the last render actually clamped it,
+	// gating the S binding and its hint.
+	summaryExpanded  bool
+	summaryScrollOff int
+	summaryTruncated bool
 
 	toast   string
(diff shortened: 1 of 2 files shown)


u9g:
Superseded — reworking this as whole-view scrolling instead of clamping the summary.

PR #899, opened 2026-07-09
fix(simulate): make the run viewer page-scrollable so long content fits short terminals

u9g:
## Problem

`lk agent simulate` / `--view <run>` rendered the run summary below the job list with no height budget. On short terminals (e.g. 31 rows) the view exceeded the screen and the header plus the top of the job list scrolled off — a 10-job run showed only jobs 7–10 at all times.

Replaces #898, which clamped just the summary; this makes the whole viewer scrollable instead.

## Fix

Compose the full page (header, description, counts, **full** job list, summary, logs) and window it to the terminal height with `↑/↓ N more lines` markers, keeping the toast and hint bar pinned at the bottom.

- Mouse wheel and PgUp/PgDn scroll the page when no pane (detail, description, logs) has focus.
- Arrow keys still move the job cursor, and the page auto-scrolls to keep the cursor row visible.
- The job list's internal two-thirds-height window is gone — the full list renders and the page scrolls instead.
- The hint bar advertises `PgUp/PgDn scroll` only when the page actually overflows.

## Testing

New `simulate_tui_layout_test.go` reproduces the report: a 31×156 terminal with 10 jobs and a long summary fits in 31 rows anchored to the top (header + Scenario 1 visible, overflow advertised); scrolling reaches the bottom of the summary with the hint pinned; cursor movement pulls the cursor row into view; content that fits renders unwindowed. Full `go test ./cmd/lk/` passes.


cmd/lk/simulate_tui_layout_test.go
theomonnom, line 1: Not sure about this file, idk if it's useful for UI stuff

theomonnom (approved):

cmd/lk/simulate_tui_layout_test.go
u9g: removed

PR #902, opened 2026-07-10
feat(auth): add support for user-based auth, public-api-server

rektdeckard:
https://linear.app/livekit/issue/DEVX-575/build-cli-poc-for-user-auth-and-list-projects
https://linear.app/livekit/issue/DEVX-577/implement-dual-mode-auth-routing-in-the-cli

- Adds experimental user-based authentication to begin implementing the new Public API service
	- Adds client codegen for the OpenAPI spec
	- Implements `lk cloud auth` flow for user sessions.
	- Implements `lk project list`, `lk project get`, `lk project create`, `lk project update`, and `lk project delete` commands using the new client behind `--experimental-auth` flag.
	- Implements net-new `lk workspace`, `lk user`, `lk simulation` commands.
	- Cache user's projects for recall with the `--project <my-project>` flag
		- On cache miss, the project cache is refreshed, meaning you don't have to synchronize anything to run commands against a new project.
	- Adds nudge to reauthenticate with new method if you have not already (and don't explicitly specify to use `--legacy-auth`)

**Architecture**

The CLI config file now stores a list of authenticated User credentials, including name, session token, and a cache of their projects. This lives alongside the existing Project credentials, and either can be used interchangeably for supported commands. Once the Public API is complete, this will be the default authentication mode for all commands.

Commands that support user-based authentication are routed through the API client in `pkg/public`. To run client codegen:

```console
go generate -tags oapigen ./pkg/public/...
```

**Usage**

A hidden global `--experimental-auth` flag gates all experimental API variants of commands. Unimplemented commands return an error. To authenticate using the new flow against staging:

```console
lk cloud auth \
  --experimental-auth \
  --server-url https://cloud-api-server-public.ochicago1a.staging.livekit.app \
  --dashboard-url https://cloud.staging.livekit.io
  ```

To run new versions of the commands:

```console
lk project list --experimental-auth
```

The default API URL can be overridden with the `--experimental-api-url` flag when running the API server locally:

```console
lk project list --experimental-auth --experimental-api-url "http://localhost:8000"
```

Note that at the moment, prod Public API Server will return "unimplemented" for all commands.


cmd/lk/project_members.go
```diff
@@ -0,0 +1,398 @@
...
+	"github.com/livekit/livekit-cli/v2/pkg/public"
+	"github.com/livekit/livekit-cli/v2/pkg/public/render"
+	"github.com/livekit/livekit-cli/v2/pkg/util"
+)
+
+// init appends the Public-API-only member and invite subcommands to `lk project`.
+// They are Hidden and require --experimental-auth (user-based auth). The project
+// they act on comes from the global --project flag (or a cached alias).
+func init() {
+	projectFlag := &cli.StringFlag{
```
Topherhindman: does having this local flag end up causing issues? something like `lk --project foo project member list` seems like it would have gripes between the local flag and the global one.

pkg/public/client.go
```diff
@@ -0,0 +1,201 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/livekit/livekit-cli/v2/pkg/public/oapi"
+)
+
+// DefaultBaseURL is the base URL of the LiveKit Public API. Override it (e.g. to
+// http://localhost:8000) for local development. The generated operation paths
+// already include the /v1 prefix, so this is the host root.
+const DefaultBaseURL = "https://beta-api.livekit.cloud"
```
Topherhindman, line 39: is this needed? seems like its a duplicate of `publicAPIBaseURL`

cmd/lk/experimental_auth.go
```diff
@@ -0,0 +1,225 @@
...
+	return cmd.Bool("legacy-auth") || (cmd.String("api-key") != "" && cmd.String("api-secret") != "" && cmd.String("url") != "")
+}
+
+// maybeShowUpgradeNotice nudges users who haven't adopted user-based auth to run
+// `lk cloud auth`. It prints once per invocation to stderr (never stdout, so it
+// can't corrupt piped data), is suppressed by --quiet, and is skipped for `cloud`
+// commands (where it would be redundant), when a session already exists, and
+// when the user explicitly opted into legacy auth (--legacy-auth or explicit
+// API-key credentials) — they've made their choice, so don't nag them.
+func maybeShowUpgradeNotice(cmd *cli.Command, conf *config.CLIConfig) {
```
Topherhindman, line 67: would this end up showing you the upgrade notice every single time if you were using the cli if you are self hosting, and have your creds set in `cli-config.yaml`?

cmd/lk/experimental_auth.go
```diff
@@ -0,0 +1,225 @@
...
+	if conf == nil || len(conf.Users) > 0 {
+		return
+	}
+	if cmd.Args().First() == "cloud" || legacyAuthForced(cmd) {
+		return
+	}
+	out.Statusf("Tip: run %s to upgrade to LiveKit's account-based API — richer access control and audit logging.", util.Accented("lk cloud auth"))
```
Topherhindman, line 77: `lk cloud auth` wouldn't actually clear the tip from being shown, since you'd have to have `--experimental-auth` included in that auth command, right? should we adjust this messaging?

cmd/lk/analytics.go
```diff
@@ -131,6 +132,10 @@ type analyticsParticipant struct {
 }
 
 func listAnalyticsSessions(ctx context.Context, cmd *cli.Command) error {
+	if experimentalAuthEnabled(cmd) {
```
Topherhindman, line 154: if we go down this path, i think we end up losing `--limit`, `--page`, `--start`, and `--end` that the old query used to support.  is this intended?

pkg/public/workspaces.go
```diff
@@ -0,0 +1,260 @@
...
+// GetWorkspace returns a single workspace by id.
+func (c *Client) GetWorkspace(ctx context.Context, workspaceID string) (*oapi.LivekitPublicapiWorkspacesV1Workspace, error) {
+	resp, err := c.gen.WorkspaceServiceGetWorkspaceWithResponse(ctx, workspaceID)
+	if err != nil {
+		return nil, err
+	}
+	if resp.JSON200 == nil {
+		return nil, responseError(resp.StatusCode(), resp.Body)
+	}
+	return resp.JSON200.Workspace, nil
```
Topherhindman: we guard json200, but return the nested optional pointer without checking it, and commands seem to dereference it (workspace.go:256, 273, 290 etc, and same pattern in members.go / invites.go / users.go).  a `200 {}` panics instead of an error. 

my beep boop recommended something like this for a generic fix that would work for all of them including the ones in client.go

```
func required[T any](p *T, what string) (*T, error) {
    if p == nil {
        return nil, fmt.Errorf("response missing %s", what)
    }
    return p, nil
}
```

cmd/lk/project_members.go
```diff
@@ -0,0 +1,398 @@
...
+	resp, err := client.InviteProjectMember(ctx, projectID, email, role)
+	if err != nil {
+		return cloudAPIError(err)
+	}
+	if cmd.Bool("json") {
+		util.PrintJSON(resp)
+		return nil
+	}
+	out.Statusf("Invited %s to project", util.Accented(email))
+	out.Statusf("Invite token: %s", util.DashString(resp.InviteToken))
```
Topherhindman: should this be `out.Result` or `out.Resultf`?

cmd/lk/cloud.go
```diff
@@ -173,6 +200,50 @@ func (a *AuthClient) ClaimCliKey(ctx context.Context) (*ClaimAccessKeyResponse,
...
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusUnauthorized {
+		// Not yet approved
+		return nil, nil
+	}
+
+	session := &ClaimCliSessionResponse{}
+	if err := json.NewDecoder(resp.Body).Decode(session); err != nil {
```
Topherhindman, line 243: should we check not just for a 200, but also that its a nonempty session token before decoding? otherwise it seems like we could persist a user without a token

cmd/lk/project.go
```diff
@@ -309,6 +355,215 @@ func listProjects(ctx context.Context, cmd *cli.Command) error {
...
+			Title(fmt.Sprintf("Delete project %s? This cannot be undone.", id)).
+			Value(&confirm).
+			WithTheme(util.FormTheme()))).
+			Run(); err != nil {
+			return err
+		}
+		if !confirm {
+			return errors.New("aborted")
+		}
+	}
```
Topherhindman: this looks like an inlined version of `confirmDestroy`.  should we just call that?

pkg/public/client.go
Topherhindman, line 1: can we add some tests for this file?

Topherhindman (commented):
overall i think this is v good. couple of preguntas

cmd/lk/project_members.go
rektdeckard: There is no legacy `lk project member list` command. This pattern of gating is only used for the commands that are ONLY implemented in public-api.

pkg/public/client.go
rektdeckard: This is actually the one to keep; `public` package should be source of truth for all this stuff. I'll get rid of the other one.

cmd/lk/experimental_auth.go
rektdeckard: Yes, but I think the expectation there would be that you always have to use `--legacy-auth`if you are self hosting...since eventually user auth will be the default. Maybe a nicer pattern is to let you set your permanent auth preference in the config and then we don't ask you again?

cmd/lk/experimental_auth.go
rektdeckard: I think this warning has to actually be switched off until user auth is the default mode, otherwise it doesn't really make sense. "Warning: you're not using the bleeding edge thing that probably doesn't work" is bad, but once it is stable, the command for people to run *will* be just `lk cloud auth`.

cmd/lk/experimental_auth.go
Topherhindman: yeah i think that would be a nicer experience.  but probably would say we do that in a separate followup pr

cmd/lk/project_members.go
rektdeckard: Kinda both? probably just the literal token should go to `stdout` though, so you can easily use result in scripts.

cmd/lk/project_members.go
rektdeckard: Sorry you're right, it's gotta just inherit the global one.

cmd/lk/project.go
rektdeckard: done

cmd/lk/cloud.go
rektdeckard: done

cmd/lk/analytics.go
rektdeckard: It was intentional, because these APIs use an opaque page cursor rather than an index. For the first pass I kinda wanted to do only the subset of things supported by legacy commands (for those that have legacy commands). But on second thought, there's no reason we can't have a different set of honored flags for each. Have added support for `--limit` and `--cursor`.

cmd/lk/experimental_auth.go
rektdeckard: done

pkg/public/client.go
rektdeckard: done

cmd/lk/project_members.go
rektdeckard: done

pkg/public/workspaces.go
rektdeckard: done

cmd/lk/experimental_auth.go
rektdeckard: turned off for now

pkg/public/client.go
rektdeckard: done

.github/workflows/build.yaml
```diff
@@ -56,12 +56,16 @@ jobs:
...
+          # prefix, so on any exact-key miss it re-restores the newest matching
+          # cache — a poisoned build cache keeps coming back until the prefix
+          # itself changes. Bump v1->v2 to force a clean rebuild across all
+          # targets. Keyed on go.sum (not cmd/pkg): the expensive webrtc/portaudio
+          # C++ objects don't depend on Go source, and Go's build cache is
+          # content-addressed, so it recompiles changed Go packages on its own.
+          key: go-v1-${{ runner.os }}-${{ matrix.id }}-${{ hashFiles('go.sum') }}
           restore-keys: |
-            go-${{ runner.os }}-${{ matrix.id }}-
+            go-v1-${{ runner.os }}-${{ matrix.id }}-
```
Topherhindman, line 68: couldn't we have just gone to the cache in GHA and deleted it? not sure we needed to change the cache key and everything, but its fine since its been done already

cmd/lk/project_members.go
```diff
@@ -0,0 +1,394 @@
...
+	resp, err := client.InviteProjectMember(ctx, projectID, email, role)
+	if err != nil {
+		return cloudAPIError(err)
+	}
+	if cmd.Bool("json") {
+		util.PrintJSON(resp)
+		return nil
+	}
+	out.Statusf("Invited %s to project with code:", util.Accented(email))
+	out.Resultf("%s\n", util.Accented(*resp.InviteToken))
```
Topherhindman: this now derefs `*resp.InviteToken` directly where `DashString` used to guard it. i ran it against a server that returns `200 {}` for the invite and it panics with a nil pointer right after printing `"Invited a@b.c to project with code:"`. `InviteProjectMember` in invites.go:61 returns `resp.JSON200` unguarded, so this is the one nested field `requirePayload` didnt get to. probably `requirePayload(resp.JSON200.InviteToken, "invite token")` in the wrapper

pkg/util/render.go
```diff
@@ -0,0 +1,115 @@
...
+	Items      []T    `json:"items"`
+	NextCursor string `json:"nextCursor,omitempty"`
+}
+
+// RenderPage prints one page of a cursor-paginated listing. As JSON it emits
+// {items, nextCursor} so a caller can iterate; as a table it prints the rows
+// and, when nextCursor is non-empty, a hint to re-run with --cursor.
+func RenderPage[T any](p *Printer, asJSON bool, list []T, nextCursor, empty string, headers []string, row func(T) []string) error {
+	if asJSON {
+		return PrintJSONTo(p.ResultWriter(), page[T]{Items: list, NextCursor: nextCursor})
```
Topherhindman, line 64: `--json` on an empty page prints `{"items": null}` because `items()` hands back a nil slice when the server omits the field, which protojson does for every empty repeated field. `jq '.items[]'` fails on it. same thing at L28 for the unpaged lists (`lk project member list --json` prints a bare null). a `if list == nil { list = []T{} }` at the top of both fixes it

.github/workflows/build.yaml
rektdeckard: Yeah, claude did this but I'm okay with it since it is self-documenting and can be "triggered" from code alone.

pkg/util/render.go
rektdeckard: Fixed here and in `RenderList`. Some legacy APIs will still exhibit `null` for empty lists, but I think preserving existing behavior there is fine.

Topherhindman: agreed

cmd/lk/project_members.go
rektdeckard: Added a guard. I am actually questioning why basically every field on these responses is a pointer type, the underlying protos don't really reflect that.

Topherhindman (approved):

cmd/lk/project_members.go
Topherhindman: careful, before you go down the spiraling rabbit hole of questioning absolutely everything, before eventually coming to question what the purpose of life is.  i went there last week. its a scary place. 😂

rektdeckard: We actually are failing to emit `required` field markers in the upstream spec file: https://github.com/livekit/public-api-server/blob/main/pkg/api/v1/openapi.yaml#L6384-L6394

Despite those fields not being optional in the proto: https://github.com/livekit/public-api-server/blob/main/protobufs/livekit/publicapi/projects/v1/projects.proto#L242-L245

rektdeckard: Will take a look at the spec generation in a follow-up

Topherhindman: oh shit. good find

PR #915, opened 2026-07-16
Add pagination support to ingress list

milos-lk:
## Summary

- Adds a `--limit` flag to `lk ingress list` and walks `NextPageToken` pages until the limit is reached or pages run out, mirroring `lk egress list` (#851).
- Pages are prepended as they arrive (each successive page is older, items within a page are oldest-first), so the final output is chronological oldest-first across page boundaries; overshoot on the last fetched page keeps its newest tail.
- Without `--limit`, behavior is unchanged: a single request returning one API page. The `--id` path is also unchanged (single request with `id` + `room` filters).
- JSON output keeps the existing `{"items": [...]}` response-object shape (unlike egress's bare array) for compatibility with existing consumers; the deprecated `list-ingress` command has no `limit` flag and keeps single-page behavior.
- Tests mirror `egress_test.go`: fake twirp Ingress service covering token walking, limit truncation, stop-on-empty-token, single-page default, and cross-page ordering.

Part of [CS-1185](https://linear.app/livekit/issue/CS-1185/implement-pagination-in-egress-and-ingress-list-apis).

## Test plan

- [ ] `go test ./cmd/lk/`
- [ ] Manual: `lk ingress list --limit N` against a project with >1 page of ingresses; confirm no overlap and chronological ordering.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


boks1971 (approved):

biglittlebigben (approved):
Thanks!

PR #917, opened 2026-07-20
feat(simulate): QOL — quota surfacing, wrapped errors, argv[0] in view hint

u9g:
- Detect inference-quota (429) exhaustion in agent logs and surface it in the TUI (dismissible dialog with a suggested `--concurrency`), plain mode, and the run report.
- Warn when `--num-simulations` is ignored because `--scenarios` is set.
- Support `--view` in non-interactive mode.
- Render all error/warning output line by line, wrapped to the terminal width, fixing lipgloss padding multiline errors to their widest line (e.g. the 348-char agent command in "agent exited before registering") and hard-wrap garbage on narrow terminals.
- Use argv[0] in the printed `agent simulate --view` re-open hint, so a renamed or path-qualified `lk` is reproduced as typed.


chenghao-mou (approved):
lgtm

theomonnom (approved):

PR #925, opened 2026-07-29
feat(simulate): export a run as JSON with --export

u9g:
Adds `--export RUN_ID` to `lk agent simulate`: prints an existing `SimulationRun` to stdout as JSON, so a run can be replayed or diffed programmatically.

```
lk agent simulate --export run_123 > run_123.json
```

- One fetch, nothing else: no agent is spawned, no run is created, and the run is never polled. A run that has not finished is an error, an unknown ID surfaces the API's not-found error, and stdout carries the full export or nothing. Flags that shape or follow a run (`--view`, `--scenarios`, `--audio`, `-n`, `--concurrency`, `--agent-name`) are rejected rather than silently dropped; `--quiet` and `--yes` still work, since the export goes to stdout and only status output is silenced.
- The run and its summary are emitted as verbatim protojson under a `{version, run, summary}` wrapper, so new proto fields reach the export on a protocol bump alone, with no change here. Fields the compiled protocol does not know are still dropped — `protojson` does not emit unknown fields — so an old `lk` against a newer server exports less than the wire carried. `summary_zstd` is dropped since the decoded summary is already included, and a summary that fails to decode fails the export instead of silently vanishing.
- Chat contexts live in the summary, as `summary.chat_history` keyed by job ID, in proto form: oneof-wrapped items, `SCREAMING_CASE` enums, RFC3339 timestamps. That is not `ChatContext.to_dict()` shape, so a replay harness needs a translation step on the way in.
- Both live runs and `--view` print the export hint. In `--view` the re-open hint is suppressed — it would just echo the command the user ran.


cmd/lk/simulate.go
```diff
@@ -98,6 +98,10 @@ var simulateCommand = &cli.Command{
 			Name:  "view",
 			Usage: "Open a pre-existing simulation",
 		},
+		&cli.BoolFlag{
+			Name:  "run-json",
```
```suggestion
			Name:  "export-json",
```

theomonnom (approved):

cmd/lk/simulate.go
theomonnom: lk agent simulate --export-json SR_123

u9g: `--output-json` or `--json` implies you can use it with the simulate flag too.

theomonnom: lk agent simulate --export SR_123

u9g: --dump-json SR_*

u9g: lk agent simulate # runs a job
lk agent simulate --dump-json # wont run a job

u9g: throw an error if the job id doesn't exist

u9g: done

theomonnom (approved):

PR #927, opened 2026-07-31
fix(agentfs): resolve a Python venv from parent directories

u9g:
## Problem

`lk agent simulate ./agent.py` in a subdirectory of a Python repo fails before the agent registers:

```
the agent exited before registering.
Command used to start agent: /Users/…/workstation/deps/bin/python3 -m livekit.agents start ./agent.py …
  /Users/…/python3: Error while finding module specification for 'livekit.agents'
  (ModuleNotFoundError: No module named 'livekit')
```

`agentfs.FindPythonBinary` looked for `.venv`/`venv` only inside the entrypoint directory. An agent under `examples/` has no environment of its own — it belongs to the repo root — and a `requirements.txt` project is detected as `python.pip`, so the `uv run --no-sync` path doesn't apply either. Resolution fell through to `LookPath("python3")`, which picks up whatever system interpreter is first on `PATH`, where `livekit-agents` is not installed.

## Fix

Delegate venv discovery to `uv python find` rather than reimplementing it. uv already:

- honors `$VIRTUAL_ENV`, and lets it outrank a nearer `.venv` on disk
- walks up to the enclosing project or workspace, so a subdirectory agent resolves the environment its repo root owns
- picks the nearest venv when several are in the ancestry

One caveat drives the only extra logic here: with no venv to find, `uv python find` succeeds and reports a *managed standalone* interpreter (`~/.local/share/uv/python/cpython-3.14-…`) that has no project dependencies installed. Accepting that would regress anyone who `pip install`ed the SDK into a system Python. A `pyvenv.cfg` beside the interpreter distinguishes a venv from a managed standalone, so only a real venv pre-empts the existing system-Python search. When `uv` isn't installed, behavior is unchanged.

Discovery runs `--offline` so it never reaches the network or triggers a Python download as a side effect, and costs ~20ms.

## Verification

Against a real monorepo checkout (`agents/examples/hotel_receptionist`, venv at the repo root), resolution goes from system `python3` to `agents/.venv/bin/python3`, and `ResolveInstalledSDKVersion` reports `1.6.6` instead of failing to import. Tests cover subdirectory resolution, `$VIRTUAL_ENV` precedence, and rejection of a non-venv interpreter; they skip when `uv` is absent.

## Note

A project with no environment of its own can now pick up a venv from an ancestor directory. That is the same tradeoff the Python toolchain makes, but it is a behavior change.

### DIFF
--- a/pkg/agentfs/installed_version.go
+++ b/pkg/agentfs/installed_version.go
@@ -42,12 +42,33 @@ func FindPythonBinary(dir string, projectType ProjectType) (string, []string, er
 		}
 	}
 
-	// Check common venv locations
-	for _, venvDir := range []string{".venv", "venv"} {
-		candidate := filepath.Join(dir, venvDir, "bin", "python")
+	// An activated venv is an explicit choice by the caller and outranks
+	// anything discovered on disk.
+	if venv := os.Getenv("VIRTUAL_ENV"); venv != "" {
+		candidate := filepath.Join(venv, "bin", "python")
 		if _, err := os.Stat(candidate); err == nil {
 			return candidate, nil, nil
 		}
+	}
+
+	// Check common venv locations, walking up so an agent in a subdirectory
+	// resolves the environment its repository root owns.
+	start := dir
+	if abs, err := filepath.Abs(dir); err == nil {
+		start = abs
+	}
+	for cur := start; ; {
+		for _, venvDir := range []string{".venv", "venv"} {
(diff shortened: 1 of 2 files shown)


chenghao-mou (approved):

theomonnom (approved):
lgtm, we could also add support to directly inspect VIRTUAL_ENV? (in case users don't have uv)

PR #935, opened 2026-08-10
Improve missing install tool guidance

### LINKED ISSUE #912: lk app create requires uv but never checks for it; the error doesn't name the missing tool
### What happened

On a clean Windows 11 machine, `lk app create --template agent-starter-python <name>` fails to install dependencies:

```
⚠  Installation failed — dependencies were NOT installed
┃ "uv": executable file not found in $PATH
┃ Fix your toolchain, then re-run the install step manually in ./speed-to-lead.
```

The command then proceeds to tell the user to run:

```
cd speed-to-lead
uv sync
uv run src\agent.py dev
```

`uv` is the only documented way to run the generated project, but the CLI:

1. never checks whether `uv` is installed before relying on it,
2. doesn't offer to install it, and
3. tells the user to *"fix your toolchain"* without naming what the toolchain is, what `uv` is, or where to get it.

### Why it matters

A developer who has never used `uv` has to stop, search, and work out that they need to `winget install --id=astral-sh.uv -e` (and then restart their shell, since it modifies PATH). That's a hard stop roughly four minutes into a first-run experience that is otherwise excellent.

The rest of the onboarding is unusually good — the starter agent runs with zero model API keys, which is a genuinely hard problem to have solved. This one missing preflight check is out of step with the rest of it.

### Environment

- Windows 11
- LiveKit CLI 2.17.0 (`winget install LiveKit.LiveKitCLI`)
- PowerShell
- Template: `agent-starter-python`

### Steps to reproduce

1. On a machine without `uv` installed
2. `winget install LiveKit.LiveKitCLI`
3. Restart shell
4. `lk cloud auth`
5. `lk app create --template agent-starter-python <name>`
6. Observe the install failure

### Suggested fix

Any one of these would close it:

- Preflight check for `uv` before template install, with a clear message naming it and linking to https://docs.astral.sh/uv/
- Offer to install it (`winget` / `brew` / `curl` depending on platform)
- At minimum, change *"Fix your toolchain"* to *"`uv` is required — install it from https://docs.astral.sh/uv/ and re-run"*

jiahao6635:
## Summary

- detect common missing-executable errors emitted by Go, env, and Unix shells
- name the missing command and explain that it must be installed and available in PATH
- preserve the existing generic guidance for unrelated installation failures
- add offline regression tests, including the Windows PATH spelling

## Testing

- go test -race ./...
- go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run
- go build -o /tmp/lk ./cmd/lk
- generated fish completion matches autocomplete/fish_autocomplete

Fixes #912

### DIFF
--- a/cmd/lk/app.go
+++ b/cmd/lk/app.go
@@ -128,6 +128,12 @@ var (
 	}
 )
 
+var missingExecutablePatterns = []*regexp.Regexp{
+	regexp.MustCompile(`(?m)["']?([a-zA-Z0-9][a-zA-Z0-9._+-]*)["']?: executable file not found in (?:\$PATH|%PATH%)`),
+	regexp.MustCompile(`(?m)^(?:env|/usr/bin/env):[ \t]+([a-zA-Z0-9][a-zA-Z0-9._+-]*):[ \t]+No such file or directory[ \t]*\r?$`),
+	regexp.MustCompile(`(?m)^(?:/bin/)?(?:sh|bash|zsh)(?:: [0-9]+)?:[ \t]+([a-zA-Z0-9][a-zA-Z0-9._+-]*):[ \t]+(?:command )?not found[ \t]*\r?$`),
+}
+
 func requireProject(ctx context.Context, cmd *cli.Command) (context.Context, error) {
 	return requireProjectWithOpts(ctx, cmd)
 }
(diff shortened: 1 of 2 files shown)


Propconnect-ai:
Ran the Windows verification on Windows 11.

**Result: could not execute the tests.** `go test ./cmd/lk/...` fails to build on Windows with the MinGW/GCC toolchain.

**Environment:** Windows 11, Go 1.26.5, GCC 16.2.0 (MSYS2 UCRT64), `CGO_ENABLED=1`

**Steps taken:** installed MSYS2 + `mingw-w64-ucrt-x86_64-gcc`, set `CGO_CXXFLAGS_ALLOW=.*` (needed for `-fms-extensions`), ran `git submodule update --init --recursive` for `pkg/portaudio/pa_src`.

**Failure:**

    platform_thread_types.cc:111:3: error: '__try' was not declared in this scope
    platform_thread_types.cc:114:5: error: '__except' was not declared in this scope

`__try`/`__except` are MSVC structured exception handling; GCC doesn't support them. Since `cmd/lk` imports `pkg/console` → `pkg/apm`, the whole package fails to build, so the new tests in `cmd/lk/app_test.go` can't run on Windows without MSVC.

Flagging because the PR adds a Windows PATH regression test — worth knowing it can currently only be exercised on Linux CI. Happy to retry with MSVC if that's the supported Windows path.

PR #937, opened 2026-08-12
simulate: give the agent under test the defaults it has deployed

u9g:
A locally spawned agent is neither hosted nor dev, so the SDK falls back to the local v1-mini turn detector and VAD interruption. Deployed, the same agent gets cloud EOT and adaptive interruption — so a simulation measures turn-taking the user will never ship: mid-readout commits on a spoken date, and every "mm-hmm" taken as a turn. The SDK gates both defaults on this one variable.


theomonnom (approved):

theomonnom:
so the models are enabled when we use LIVEKIT_DEV_MODE?

u9g:
> so the models are enabled when we use LIVEKIT_DEV_MODE?

the EOT and barge-in, yeah.

theomonnom (approved):

PR #938, opened 2026-08-13
feat(agents): implement cross-region deployment confirmation

rektdeckard:
https://linear.app/livekit/issue/DEVX-636/warn-in-cli-when-eu-agent-deploy-uses-us-observability-region

Show warning and require additional confirmation when non-EU project elects to deploy an agent to an EU cluster.

https://github.com/user-attachments/assets/dadf5b8b-0b33-48da-a04d-fd49340e750b

RELATED: https://github.com/livekit/cloud-agents/pull/780

### DIFF
--- a/cmd/lk/agent.go
+++ b/cmd/lk/agent.go
@@ -1967,7 +1967,7 @@ func resolveRegion(cmd *cli.Command, settingsMap map[string]string, title string
 	for _, r := range regionOptions {
 		label := r
 		if slices.Contains(warnRegions, r) {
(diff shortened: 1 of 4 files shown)


real-danm (approved):
is this a breaking change with the current flow? if it is release will need to be major version bump which would need to be reflected in the server-sdk-go as well.

Clients will need to update before they see this message, i think we can enforce this on the config side for the min cli version. that's for a future PR

rektdeckard:
The change is backwards-compatible, we fall back to the default flow if the new client settings fields are not present. Added some tests to verify.

PR #949, opened 2026-08-27
fix(cli): Fix project add not respecting --default flag

ChelseaBradbury:
When creating a project, I noticed that the `--default` flag is not respected unless the project name is passed as an argument to the command. It tries to set the default project to the project's name, but it hasn't prompted for a project name yet, so it sets it to empty string.

This fix produces what I assume is the intended behavior.


rektdeckard (approved):
Thanks @ChelseaBradbury, nice catch

PR #951, opened 2026-08-28
ci: one cross-build runner per target, one target per OS on PRs

u9g:
The `Cross-build linux+windows (zig)` job takes ~19 minutes, ~17 of which are a single `GoReleaser snapshot build` step. It compiles all four zig cross targets in one goreleaser invocation on one `ubuntu-latest` (4 vCPU) runner:

- 445 `.cc`/`.c` under `pkg/apm` + 114 under `pkg/portaudio` ≈ 560 translation units
- × 4 targets = ~2,240 C++ compiles at `-O2` through `zig cc`
- 17 min × 4 cores ÷ 4 targets ≈ 1.8 s/TU — a cold WebRTC APM build, every run

Nothing is reused between runs either: `setup-go`'s cache key is an exact `go.sum` hash with no `restore-keys`, and even when it hits (699 MB restored in [run 33115504417](https://github.com/livekit/livekit-cli/actions/runs/33115504417)) the step still took 19m19s vs 19m45s cold.

This PR does the cheap structural half:

1. **Matrix the four targets** so each gets its own runner instead of sharing 4 vCPUs. Same total CPU, ~4× less wall clock.
2. **Build one target per OS on PRs** (`lk-linux-amd64`, `lk-windows-amd64`); pushes to `main` still build all four, so nothing reaches a release untested.

The `.cross` cache key gains the target id, since each runner now populates only its own `.cross/<target>/`.

Expected: ~19 min → ~5 min on PRs.

Not addressed here: why a warm 699 MB `GOCACHE` buys nothing. Worth a follow-up.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -32,18 +32,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
-      # PRs build one target per OS; main covers all four before release.
-      # fromJSON is required, not decorative: a matrix value must be an array,
-      # and the expression language has no array literal — only boolean, null,
-      # number and string — so a list that varies by event has to arrive as
-      # parsed JSON.
       matrix:
-        id: |-
-          ${{ case(
-            github.event_name == 'pull_request',
-            fromJSON('["lk-linux-amd64", "lk-windows-amd64"]'),
-            fromJSON('["lk-linux-amd64", "lk-linux-arm64", "lk-windows-amd64", "lk-windows-arm64"]')
-          ) }}
+        id: [lk-linux-amd64, lk-linux-arm64, lk-windows-amd64, lk-windows-arm64]
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -24,10 +24,26 @@ permissions:
...
+      # fromJSON is required, not decorative: a matrix value must be an array,
+      # and the expression language has no array literal — only boolean, null,
+      # number and string — so a list that varies by event has to arrive as
+      # parsed JSON.
+      matrix:
+        id: |-
+          ${{ case(
+            github.event_name == 'pull_request',
+            fromJSON('["lk-linux-amd64", "lk-windows-amd64"]'),
+            fromJSON('["lk-linux-amd64", "lk-linux-arm64", "lk-windows-amd64", "lk-windows-arm64"]')
```
rektdeckard: ARM builds only happen on merge to `main` then? I suppose that's OK since these take the lion's share of the time on PRs, but we do lose some confidence.

rektdeckard (approved):
Looks good, I think we can live with ARM build tests only on `main`

.github/workflows/build.yaml
u9g: You're right, I undid this change.

PR #956, opened 2026-09-01
fix(install): retry GitHub downloads with exponential backoff

stephen-derosa:
The installer served at https://get.livekit.io/cli makes three unguarded curl calls to GitHub (the releases API, the archive, and checksums.txt). Any transient failure aborts the whole install, which makes CI runners flaky.

Retry each curl command, overridable for custom setting.

Also adds the missing `jq` prerequisite check — the script already depended on jq but only reported it as an unhelpful version parse error.

### DIFF
--- a/install-cli.sh
+++ b/install-cli.sh
@@ -31,10 +31,20 @@ FISH_COMPLETION_PATH="/usr/share/fish/vendor_completions.d"
 log()   { printf "%b\n" "$*"; }
 abort() { printf "%s\n" "$@" >&2; exit 1; }
 
+
+# GitHub is frequently flaky , so every download is retried.
+# Overridable for testing / impatient callers.
+MAX_ATTEMPTS="${LK_INSTALL_MAX_ATTEMPTS:-5}"
+RETRY_DELAY="${LK_INSTALL_RETRY_DELAY:-2}"
+
+[[ "$MAX_ATTEMPTS" =~ ^[0-9]+$ ]] && [ "$MAX_ATTEMPTS" -ge 1 ] || abort "LK_INSTALL_MAX_ATTEMPTS must be a positive integer"
+[[ "$RETRY_DELAY" =~ ^[0-9]+$ ]] && [ "$RETRY_DELAY" -ge 1 ] || abort "LK_INSTALL_RETRY_DELAY must be a positive integer (seconds)"
+
 [ -n "${BASH_VERSION:-}" ] || abort "This script requires bash"
 [ -d "$INSTALL_PATH" ]     || abort "Could not install, $INSTALL_PATH doesn't exist"
 command -v curl >/dev/null || abort "cURL is required and is not found"
 command -v sha256sum >/dev/null || abort "sha256sum is required and is not found"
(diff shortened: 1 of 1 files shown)


install-cli.sh
```diff
@@ -31,10 +31,20 @@ FISH_COMPLETION_PATH="/usr/share/fish/vendor_completions.d"
 log()   { printf "%b\n" "$*"; }
 abort() { printf "%s\n" "$@" >&2; exit 1; }
 
+
+# GitHub is frequently flaky , so every download is retried.
+# Overridable for testing / impatient callers.
+MAX_ATTEMPTS="${LK_INSTALL_MAX_ATTEMPTS:-5}"
```
alan-george-lk, line 37: Should we default to 0 to maintain current script behavior?

Also should we consider script args in addition to env vars? Or is this just intended for us internally? There's no `install-cli.sh -h` that lists these out to console for anyone to see

alan-george-lk (approved):

install-cli.sh
stephen-derosa: No i dont think we should keep the current behavior since it causes false failures due GH being unreliable.

im not positive the design motivation of the script so i didnt want to bloat with taking in args.

install-cli.sh
```diff
@@ -31,10 +31,20 @@ FISH_COMPLETION_PATH="/usr/share/fish/vendor_completions.d"
...
+RETRY_DELAY="${LK_INSTALL_RETRY_DELAY:-2}"
+
+[[ "$MAX_ATTEMPTS" =~ ^[0-9]+$ ]] && [ "$MAX_ATTEMPTS" -ge 1 ] || abort "LK_INSTALL_MAX_ATTEMPTS must be a positive integer"
```
u9g, line 40: is this a real issue that we think someone would run into or just overprotective tests?

u9g (dismissed):
LGTM, not sure about text/negative max attemps or retry delay guards make sense though.

install-cli.sh
```diff
@@ -31,10 +31,20 @@ FISH_COMPLETION_PATH="/usr/share/fish/vendor_completions.d"
 log()   { printf "%b\n" "$*"; }
 abort() { printf "%s\n" "$@" >&2; exit 1; }
 
+
+# GitHub is frequently flaky , so every download is retried.
+# Overridable for testing / impatient callers.
+MAX_ATTEMPTS="${LK_INSTALL_MAX_ATTEMPTS:-5}"
+RETRY_DELAY="${LK_INSTALL_RETRY_DELAY:-2}"
```
u9g, line 38: Do we have defaults for these anywhere? if so, I don't see it.

install-cli.sh
stephen-derosa: over protective tests in case users have invalid env vars

install-cli.sh
stephen-derosa: the default is the `:*` so `5` and `2` respectively

u9g (approved):

rektdeckard (approved):

PR #968, opened 2026-09-08
simulate: list, view and export subcommands

u9g:
Requested in Slack: `lk agent simulate --view` can reopen a past run, but there was no way to list runs to find an ID.

Rather than add `--list` next to `--view` and `--export`, this makes the run operations subcommands. Each of those flags took an ID and replaced the whole run action, and `audio` inherited them, so `lk agent simulate audio --view ID` parsed and did something meaningless. A flag that swaps the action is a subcommand; this matches the noun/verb shape of the rest of `lk` (`lk agent list`, `lk room list`, `lk egress list`) and the `git stash` / `git stash list` / `git stash show` precedent of a verb whose bare form does the main thing.

```
lk agent simulate                     prints the subcommand list, runs nothing (was: text run)
lk agent simulate text [entrypoint]   run, text mode (new; was bare simulate)
lk agent simulate audio [entrypoint]  run, audio pipeline (unchanged)
lk agent simulate list                newest runs for the project (new)
lk agent simulate view <run-id>       was --view
lk agent simulate export <run-id>     was --export
```

**Commit 1** turns `--view`/`--export` into `view`/`export` with a positional run ID and updates the exit hints to print the new forms. The old flags are removed, not aliased: the hint is reprinted after every run, so only stale scrollback breaks.

**Commit 2** adds `list`: one `ListSimulationRuns` call for the project's first page (server returns newest first, 20 per page, no jobs), rendered like `lk agent list` with `--json` for the raw response.

**Commit 3** adds `--limit` to `list`, the same `next_page_token` loop as `lk egress list`: unset keeps the single-page default, `--limit N` walks pages until N runs are collected. `--json` prints the collected runs inside the List response so the field formatting is unchanged. No filters yet.

**Commit 4** adds `text`: `simulate audio` names its mode, so `simulate text` should exist too.

**Commit 5** makes bare `simulate` a help command. With `list`/`view`/`export` under it, a verb that also runs on its own is the odd one out, and `lk agent simulate` typed alone now shows the options instead of starting a run. The project is not resolved on the help path. Docs that say `lk agent simulate` for a text run need to say `lk agent simulate text`.

The run flags on `simulate` stay persistent, so `list`/`view`/`export` still parse `--scenarios` etc. and ignore them. Making them local would break `lk agent simulate --scenarios f audio`: urfave/cli v3 resolves a flag by the first command in the lineage that declares it and re-runs `PreParse` on the child's flags, so a shared or duplicated flag loses the value parsed on the parent.

Tested against my dev project: `list`, `list --limit 3`, `list --json`, `view <id>` (CI and TUI paths), `export <id>`. Paging tested against a staging project with 184 runs: `--limit 45` returned 45 unique runs across three pages, newest first with no seam; `--limit 1000` drained all 184.


theomonnom (approved):

u9g:
Demo of the subcommands against my dev project: bare `lk agent simulate` printing help and running nothing, `text --help`, `list`, `list --limit 3`, `list --json`, `export`, `view` in CI mode and in the TUI, and `audio --view` failing to parse.

https://github.com/user-attachments/assets/30d37827-3d9b-4bbc-8a7e-3fb62807c8eb

theomonnom (approved):
lgtm!

PR #970, opened 2026-09-08
feat(simulate): require stable ids in scenarios.yaml, offer to add them

u9g:
Depends on livekit/protocol#1785 (adds `Scenario.id` and `ScenarioGroup.id`). go.mod is pinned to that branch's commit so CI builds; bump to the merged version before merging this.

## Why
Today nothing correlates runs of the same scenario over time: the label, instructions, and agent expectations can all be edited, and hashing the content would make every edit a new identity. Dashboards charting a scenario's success rate over time need a stable key.

## What
- `scenarios.yaml` now requires an `id` on the file and on every scenario. Loading rejects missing or duplicate ids.
- A file missing ids is refused. Interactively the CLI offers to insert generated 8-character ids in place, as the first key of each entry, preserving comments, key order, and indentation (yaml.v3 node edit, no re-marshal). `--yes` inserts without asking; non-interactive without `--yes` errors with instructions.
- Scenarios rendered by the CLI (generated runs saved to temp or via the save key, clipboard copy) are given ids so the resulting file runs as-is.
- Ids are sent on `ScenarioGroup` / `Scenario` when creating the run. Run-level identity already exists as `SimulationRun.id`.

## Example
```yaml
id: 3k9fz2ab
name: drive-thru
scenarios:
  - id: q7x1m0dp
    label: order a burger
    instructions: ...
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


u9g:
Ids now come from protocol's shared guid utility with stable prefixes: `SCN_` for scenarios and `SCNG_` for the scenario group (livekit/protocol#1785). Dropped the `randomBase36` refactor commit since it is no longer needed.

theomonnom (approved):

PR #973, opened 2026-09-11
Show a remote banner fetched from banner.json on main

u9g:
## Why

We want to tell people running an installed `lk` things like "a new version is out, run `brew upgrade livekit-cli`" or "livekit.toml layout changed" without shipping an auto-updater (decided against) and without waiting for them to update to see the notice.

## What

- `banner.json` at the repo root, fetched from `main` on every run. It is a list of notices, each with its own version selector:
  ```json
  [
    {"message": "lk agent simulate can now export runs. Upgrade: brew upgrade livekit-cli", "versions": "< 2.18.0"},
    {"message": "lk 3.0 is out and moves the agent id to [cloud] in livekit.toml.", "versions": "< 3.0.0"}
  ]
  ```
  `versions` is a semver constraint; empty matches every build. Every entry that matches prints, in file order, so a new-feature announcement for old builds can sit next to an upgrade notice. The committed file is `[]`. Editing the file on `main` is the whole publishing process.
- Shown at the top, before the command's output, inside a fence on stderr via `out.Status`, so it honors `--quiet` and never mixes with stdout. Only shown on an interactive terminal, so scripts and pipes are unaffected.
- The fetch starts during startup and runs concurrently with the command. Output written through the printer is held until the fetch completes, then flushed behind the banner, so the banner stays at the top and the round trip hides behind the command's own RPCs. The hold is bounded by a 1s timeout, and the file is cached at `~/.livekit/banner.json` for an hour, so only the first run each hour fetches at all (~70-170ms to raw.githubusercontent.com). A failed fetch falls back to the stale cache. The root `After` hook waits for the flush so a fast command cannot exit with output still held. Non-interactive runs skip the fetch entirely.

## Verification

`go build ./cmd/lk`, `go vet ./cmd/lk`, `go test ./cmd/lk` pass. Demo video in a comment below.


u9g:
Demo: `banner.json` as a list of two notices (served locally for the recording). A 2.17.0 build prints both notices fenced at the top of `lk room list` on stderr, on the very first run. 2.18.6 matches only the `< 3.0.0` one, `--quiet` silences it, `--json | head` keeps stdout clean, and a 3.0.0 build runs clean.

https://github.com/user-attachments/assets/00e80a89-2cf0-48f2-9866-2696c5cd253c

cmd/lk/banner.go
```diff
@@ -0,0 +1,165 @@
...
+)
+
+// banner.json on main is the list of notices shown to installed CLIs. Editing that
+// file is the whole release process: no build or tag involved.
+const bannerURL = "https://raw.githubusercontent.com/livekit/livekit-cli/main/banner.json"
+
+// The banner prints before the command, so the fetch is awaited. To keep that off
+// most runs, the fetched file is cached and reused for bannerTTL.
+const (
+	bannerTimeout = time.Second
```
rektdeckard, line 46: Is this enough time for remote locations / mobile hotspot / etc?

cmd/lk/banner.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	// The fence sets the notices apart from the command's own output. The fixed
+	// width wraps long messages instead of letting the border break on narrow
+	// terminals.
+	fence := lipgloss.NewStyle().
+		Border(lipgloss.RoundedBorder()).
+		BorderForeground(util.Warning()).
+		Padding(0, 1).
+		Width(76)
+	out.Statusf("%s\n", fence.Render(strings.Join(msgs, "\n\n")))
```
rektdeckard: Maybe put a separator between message?

cmd/lk/banner.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return nil
+	}
+	var msgs []string
+	for _, b := range banners {
+		if b.Message == "" {
+			continue
+		}
+		if b.Versions != "" {
+			c, err := semver.NewConstraint(b.Versions)
```
rektdeckard, line 128: Nice, I was gonna check you were using this since we had it already

rektdeckard (approved):
This looks great, small nits mainly. My one question is whether there will ever be other conditions besides just your lk version that might want to trigger a banner? A specific command for example might want to provide usage tips that aren't dependent on version, but aren't appropriate to show at all times.

cmd/lk/banner.go
u9g: Done in 5b2dd16: a dimmed horizontal rule now sits between notices inside the fence.

u9g:
Demo of the separator: one notice alone, then three entries in banner.json of which two match this version and render with a rule between them, then a piped run that skips the banner.

https://github.com/user-attachments/assets/f6f701f2-bc61-4ec5-ab49-79835c79b87f

cmd/lk/banner.go
u9g: Video of how it renders: https://github.com/livekit/livekit-cli/pull/973#issuecomment-5665391626

cmd/lk/banner.go
u9g: The fetch now starts at launch and runs alongside the command, with output held until it lands, so for anything that makes an RPC the round trip is hidden entirely. Worst case is still a 1s pause once an hour, and the banner just waits for the next run.

rektdeckard (approved):

rektdeckard:
LGTM! Just reiterating the question: do we want to add an optional "command-level" filter so that banners can be shown only for specific commands and their subcommands? Also happy to cross that bridge when we get there.

u9g:
> LGTM! Just reiterating the question: do we want to add an optional "command-level" filter so that banners can be shown only for specific commands and their subcommands? Also happy to cross that bridge when we get there.

Let's cross that bridge when we get there

PR #978, opened 2026-09-15
agent: [cloud] id block and multi-region livekit.toml

u9g:
## Why

One `livekit.toml` per agent should hold the agent's name and every region it deploys to. Today the id sits under `[agent]` next to the name, and deploying to two regions means two toml files and two deploy invocations. Proposal: https://app.notion.com/p/livekit/Proposal-livekit-toml-layout-for-agent-name-and-regions-3d83c4901a4281beb3b9f1f3e74de44d

Stacked on #972, which adds `[agent] name`.

## What

**Layout.** The Cloud Agents id moves out of `[agent]` into a `[cloud]` block. A single-region agent:

```toml
[agent]
name = "my-agent"

[cloud]
id = "CA_xxxxxxxxxxxx"
```

An agent deployed per region, one `[cloud.<region>]` table each, region codes as `--region` accepts them:

```toml
[agent]
name = "my-agent"

[cloud.us-east]
id = "CA_aaaaaaaaaaaa"

[cloud.eu-central]
id = "CA_bbbbbbbbbbbb"
```

`id` lives in `[cloud]` or in the region tables, never both. A legacy `[agent] id` is still read and migrates to `[cloud]` the next time the file is saved.

**Commands.**
- `lk agent create` and `lk agent config` write `[cloud] id`. `lk agent create --region <r>` into a file that already has an agent adds a `[cloud.<r>]` entry (converting a flat id to its region table by looking the region up) instead of replacing the agent. Cloud assigns the name; if it differs from `[agent] name` the CLI warns and keeps the file's name.
- `lk agent deploy` deploys every region the file lists, or only the one passed with `--region`.
- `status`, `update`, `restart`, `rollback`, `logs`, `delete`, `versions`, `secrets` prompt for a region when the file lists several (non-interactive runs need `--id`).

No server or protocol changes.

## Verification

`go build ./cmd/lk`, `go vet ./...`, `go test ./pkg/config ./cmd/lk` pass. New `pkg/config/livekit_test.go` covers legacy id migration, region tables, id/region exclusivity, and the saved layout.


u9g:
Demo against the agent-test-arena project: a legacy `[agent] id` still resolves, `lk agent config` rewrites the file with `[cloud] id`, and with two `[cloud.<region>]` tables `deploy --region` rejects an unlisted region, `--yes` runs ask for `--id`, and interactive runs prompt for the region.

https://github.com/user-attachments/assets/47f8ba11-f6c9-4cda-a6e6-bc658a951ae4

PR #982, opened 2026-09-16
feat(simulate): default to ./scenarios.yaml

u9g:
Split out of #969 (1 of 2 here; the save prompt is stacked on top of this one).

## Summary

- `lk agent simulate` (and `simulate text` / `simulate audio`) picks up `./scenarios.yaml` when `--scenarios` is omitted and that file exists. So a project that keeps its scenarios checked in stops re-passing the flag.
- `--agent-name` is satisfied by the default file too; its error now names both ways to provide one.
- An explicit `--scenarios` still wins and is never second-guessed: a missing path is passed through so the existing "failed to read scenarios file" error fires instead of silently generating from source.
- With no file at all, nothing changes — scenarios are still generated from the agent's source.

## Test plan

- [x] `go test ./cmd/lk/...` passes; new `TestScenariosPathOrDefault` covers default pickup, no-file, explicit-flag-wins, and explicit-but-missing
- [x] Fish completions regenerated (`go generate ./cmd/lk`)


u9g:
The bakery agent project, which keeps its scenarios in `scenarios.yaml`: the `--agent-name` error names both ways to provide a scenarios file, and `lk agent simulate text` with no `--scenarios` loads the checked-in file ("Loaded 3 scenarios from scenarios.yaml") and runs it.

https://github.com/user-attachments/assets/9e43f351-df8a-4a1c-8dd6-1ebd16b997f8

PR #983, opened 2026-09-16
feat(simulate): offer to save generated scenarios to scenarios.yaml

u9g:
Split out of #969. Stacked on #982 (which adds the `scenarios.yaml` default this writes to) — review that one first; the diff here is TUI-only.

## Summary

- When a source-generated run's scenarios first arrive, the TUI asks once whether to write them to `scenarios.yaml`. With #982, the next run in that project picks the file up with no flag.
- Accepting when that name is already taken opens the existing file-name dialog instead of overwriting; the file on disk is left untouched.
- `esc` / `n` declines, and the `s` key still saves later. Declining is remembered for the rest of the run, so the offer never re-appears.

## Test plan

- [x] `go test ./cmd/lk/...` passes
- [x] New VRT frame `simulate_save_offer`; no other golden frames moved
- [x] Key-dispatch cases for `esc` and `tab`, plus tests for the accept path and the name-conflict fallback


u9g:
A project with no scenarios file: the run generates scenarios from the agent's source, the offer appears as soon as they arrive, Save writes `scenarios.yaml`, and `head` shows the file that the next run picks up (with #982).

https://github.com/user-attachments/assets/b92c0511-2717-414a-9f91-3178c69eacf4

theomonnom (approved):

PR #985, opened 2026-09-22
simulate: send staging projects to the staging cloud API

u9g:
## Problem

`lk agent simulate list` (and every other simulate subcommand) sends its RPCs to `https://cloud-api.livekit.io` regardless of which project is selected. With a staging project as the default, that fails:

```
$ lk agent simulate list
Using project [jasons-staging-development-proj]
unable to list simulation runs: twirp error unauthenticated: invalid credentials
```

The only workaround was the hidden `--server-url` flag, and the comment on `dashboardBaseURL` pointed at `cloud-api.staging.livekit.io`, which is a Vercel site that 404s for every route.

## Fix

Derive the cloud API host from the project URL in the `simulate` Before hook, using the same staging/production match `cloudProject` already uses for dashboard links. `--server-url` still wins when set.

Only the simulate subcommands change. The other `lk agent` commands share the same `serverURL` default and are left as they were.

## Testing

- New table test for `cloudAPIURL`.
- Staging project, no flags: lists runs.
- Prod project, no flags: unchanged.
- `--server-url https://cloud-api.livekit.io` against the staging project still reproduces the original error, confirming the override is respected.


u9g:
Staging project (default) lists its runs with no `--server-url`, and the printed re-open hint no longer carries the derived host; a production project still goes to `cloud-api.livekit.io` unchanged.

https://github.com/user-attachments/assets/ff576a68-aff5-4d02-b024-a96830af0d27

theomonnom (approved):

